PR #13, opened 2015-03-25
more explicit update instructions

waldyrious:
mention full command

### DIFF
--- a/lib/messages.js
+++ b/lib/messages.js
@@ -7,6 +7,6 @@ exports.emptyCache = function() {
 
 exports.notFound = function() {
   return 'Page not found\n' +
-  'Try updating with tldr --update, or submit a pull request to:\n' +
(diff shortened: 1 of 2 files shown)


lib/messages.js
```diff
@@ -7,6 +7,6 @@ exports.emptyCache = function() {
 
 exports.notFound = function() {
   return 'Page not found\n' +
-  'Try updating with --update, or submit a pull request to \n' +
+  'Try updating with tldr --update, or submit a pull request to:\n' +
```
rprieto: Good idea, maybe even `"tldr --update"` (with quotes) to show that it's a command?

waldyrious: Sure. Shall I do that in both cases?

rprieto:
Nice one :+1:

Issue #12, opened 2015-03-25
Update fails (invalid signature)


rubenvereecken:
Just copying the issue from tldr-pages/tldr#263. 

Related error:
```
➜  ~  tldr --update
Updating...
Error: invalid signature: 0x2e736370
    at /usr/local/lib/node_modules/tldr/node_modules/unzip/lib/parse.js:59:13
    at processImmediate [as _immediateCallback] (timers.js:345:15)
```

A lot of people are getting them it seems, including a friend I introduced to the client. It may also be the cause of #11 because the cache got filled improperly.

rprieto:
That's a weird one. Also happened to me, then went away on its own. Could it be due to failed downloads, and we try to `unzip` the file anyway?

Otherwise, there are mentions of [the unzip module failing if the archive contains files with "store"-level compression](https://github.com/EvanOxfeld/node-unzip/issues/32). We could try https://www.npmjs.com/package/unzip2.

adius:
I get the same error. Is there a quickfix to get it working again?

rubenvereecken:
It should actually already be fixed in the master branch. Try pulling that
in instead of installing it through NPM :).

If that doesn't work I've seriously overlooked something.

adius:
…and why don't you release it on npm? ^^

rubenvereecken:
Good point, @rprieto let's update, the current release op npm is seriously broken :smile:.

kud:
I've got the same problem. Please, publish on npm.

rubenvereecken:
@rprieto publish :+1:

rprieto:
Done, as `1.0.1`! Sorry for the super slow publish. @rubenvereecken let's figure out how to add you as a publisher on npm :)

kud:
```console
~/P/_/project ❯❯❯ npm install -g tldr                                        ⏎  ⬆
/Users/kud/.nvm/versions/node/v0.12.4/bin/tldr -> /Users/kud/.nvm/versions/node/v0.12.4/lib/node_modules/tldr/bin/tldr
tldr@1.0.1 /Users/kud/.nvm/versions/node/v0.12.4/lib/node_modules/tldr
├── commander@2.3.0
├── deep-extend@0.2.11
├── ms@0.6.2
├── rimraf@2.2.8
├── wrench@1.5.8
├── async@0.9.2
├── marked@0.3.3
├── colors@1.0.3
├── cpr@0.3.3 (graceful-fs@3.0.8)
[...]

PR #14, opened 2015-04-09
Installed unzip2 to handle STORE zip compression

rubenvereecken:
As @rprieto hinted in #12 `unzip` fails at opening zip files using the `Store` compression method. Installing `unzip2` solves this as it explicitly supports this compression method. 

All tests pass and this resolves issue #12.


rubenvereecken:
I noticed npm reordered some packages in `package.json` because I used its `--save` option. Let me know if I have to repair the old order.

rprieto:
Nice work @rubenvereecken ! Let's merge this in.
IMO sorting the package is a good thing, I think `npm` starting sorting recently and we hadn't `--save`ed any packages since.

PR #22, opened 2015-12-04
Fix #15: upgrade commander to 2.9.0

### LINKED ISSUE #15: `tldr sort` tries to find the `tldr-sort` command
This was extremely weird. I didn't bother to even try and find the source of this strange bug but I might later on.

`tldr sort` threw me the following error after correctly displaying the page for sort:
```
module.js:340
    throw err;
          ^
Error: Cannot find module '/usr/local/bin/tldr-sort'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```

This is just a bit silly, right? I tried `tldr ls` and it didn't have the problem.

igorshubovych:
The problem is related to bug in `commander` package
https://github.com/tj/commander.js/issues/248

It was fixed in 2.4.0 of `commander`. See their [History](https://github.com/tj/commander.js/blob/master/History.md#240--2014-10-17) for more details.

But I updated the package to 2.9.0, because more than a year came after 2.4.0 was released and there are many bug fixes introduced since then.

I tested it under Node 0.12,  4.1 and 5.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   "dependencies": {
     "async": "~0.9.0",
     "colors": "~1.0.1",
-    "commander": "~2.3.0",
+    "commander": "~2.9.0",
     "cpr": "~0.3.2",
     "deep-extend": "~0.2.8",
     "lodash": "~2.4.1",


rprieto:
Thanks @igorshubovych, nice one :+1:

PR #24, opened 2015-12-07
Updating npm packages

igorshubovych:
Let's stay up to date

Not old

I tested these commands together with unit tests.

~~~shell
NODE_ENV=development node ./bin/tldr zip && sleep 2 && \
NODE_ENV=development node ./bin/tldr --update && \
NODE_ENV=development node ./bin/tldr git && sleep 2 && \
NODE_ENV=development node ./bin/tldr --clear-cache && \
NODE_ENV=development node ./bin/tldr --update && \
NODE_ENV=development node ./bin/tldr tar && \
NODE_ENV=development node ./bin/tldr git && \
NODE_ENV=development node ./bin/tldr --render test.md
~~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,24 +30,24 @@
     "!win32"
   ],
   "dependencies": {
-    "async": "~0.9.0",
-    "colors": "~1.0.1",
+    "async": "~1.5.0",
+    "colors": "~1.1.2",
     "commander": "~2.9.0",
-    "cpr": "~0.3.2",
-    "deep-extend": "~0.2.8",
-    "lodash": "~2.4.1",
+    "cpr": "~1.0.0",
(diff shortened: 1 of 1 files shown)


igorshubovych:
@rprieto what do you think about it?

When I was testing this, I upgraded one package, then run full tests and tested manually all commands, only then I started to upgrading the next packages. And I did this for each package from dependencies and devDependencies.

rprieto:
Thanks a lot for that, you're right we should really try and keep them up to date.
Also :+1: for the best relevant gif!

*ps:* the manual integration test is a good idea, maybe we can think of ways to automate this as part of the test suite?

igorshubovych:
More tests. I plan to have 100% test coverage, or close to it.

PR #26, opened 2015-12-13
Correctly process Unix redirect characters

### LINKED ISSUE #25: Character "<" (less) is rendered as "≪"
Consider the following page

~~~Markdown
# bc

> Calculator.

- Calculate the result of expression

`bc <<< "(1 + 2) * 2 ^ 2"`
~~~

It is rendered
```
  bc
  Calculator.

  - Calculate the result of expression
    bc ≪≪≪ "(1 + 2) * 2 ^ 2"
```

It is wrong and cannot be safely copy-pasted to run in CLI.

igorshubovych:
This should fix #25.

I used [string.js](https://www.npmjs.com/package/string) instead of [unhtml](https://www.npmjs.com/package/unhtml). It is much better supported and maintained.

I also added tests for this.

Probably it is better in future to refactor unhtml function in its own module and test it separately.


rprieto:
Nice fix, and thanks for adding the corresponding test! :+1:

PR #27, opened 2015-12-14
Display several code lines per one example

igorshubovych:
Try `tldr cal`.

Before
```
  cal
  Prints calendar information

  - Display a calendar for the current month or specified month
    cal -m Dec

  - Display a calendar for the current year or a specified year
    cal 2013

  - Display date of Easter (western churches)
    ncal -e 2013
```

After
```
  cal
  Prints calendar information

  - Display a calendar for the current month or specified month
    cal
    cal -m 12
    cal -m Dec

  - Display a calendar for the current year or a specified year
    cal -y
    cal 2013

  - Display date of Easter (western churches)
    ncal -e
    ncal -e 2013
```

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -28,7 +28,7 @@ exports.parse = function(markdown) {
     if (/^<code>.*<\/code>$/.test(text)) {
       var example = _.last(page.examples);
       if (example) {
-        example.code = unhtml(text);
+        example.code.push(unhtml(text));
       }
     } else {
       return text;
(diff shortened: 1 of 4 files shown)


igorshubovych:
See also
https://github.com/tldr-pages/tldr/pull/313

rprieto:
Thanks @igorshubovych. This is an interesting one because it's against the TLDR guidelines, which say we should have 1 command per example, so we should really fix these pages. Do you still think it's worth handling until the outliers are fixed?

igorshubovych:
No. Probably not.

I will try to fix the pages.

PR #30, opened 2015-12-20
Undeprecate overriding the operating system

igorshubovych:
Consider the following use cases:

- User works as a system administrator on Windows or OSX machine, but they are managing Linux servers. They cannot install tldr everywhere, so they run it on their machine but they want to search command syntax for target platform.
- User is using his mobile device - Android or iOS (there is no iOS client yet, but it can be created soon with existing code and PhoneGap technology). They want to check command syntax for their Linux/OSX laptop.

Using this patch they can configure platform in `$HOME/.tldrrc`:

~~~json
{
  "platform": "linux"
}
~~~

Or in command line:

`tldr du --os=linux`

### DIFF
--- a/lib/platform.js
+++ b/lib/platform.js
@@ -1,18 +1,39 @@
 var os      = require('os');
+var config  = require('./config');
 
 var folders = {
-  darwin: 'osx',
-  linux : 'linux',
-  sunos : 'sunos'
+  'osx': 'osx',
+  'darwin': 'osx',
+  'linux': 'linux',
+  'sunos': 'sunos'
 };
 
-exports.resolve = function(command) {
-  return [
-    specific(command, folders[os.platform()]),
-    specific(command, 'common')
-  ];
-};
+function isSupported(platform) {
+  return folders.hasOwnProperty(platform);
+}
+
+function getPreferredPlatform() {
+  var platform = config.get().platform;
+  if (isSupported(platform)) {
+    return platform;
+  }
+  return os.platform();
+}
 
 function specific(command, os) {
   return 'pages/' + os + '/' + command + '.md';
 }
+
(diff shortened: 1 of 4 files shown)


lib/platform.js
```diff
@@ -1,14 +1,26 @@
...
 };
 
+exports.isSupported = function isValid(platform) {
+  // return ['osx', 'linux', 'sunos'].indexOf(platform) >= 0;
+  return folders.hasOwnProperty(platform);
+};
+
+exports.getPrefered = function getPrefered() {
+  var configuredPlaform = config.get().platform;
+  return configuredPlaform || folders[os.platform()];
```
rprieto: From what I understand, this function returns the name of the folder for the current platform. Can we rename is to `folderForPreferredPlatform` for example?

Also, should we use `folders[configuredPlaform]` so that users can specify aliases, like `--os=darwin` instead of having to specify the real folder name?

rprieto:
A few comments here and there, but a great idea to keep the overrides! Nice work @igorshubovych.

igorshubovych:
I overwrote the commit.

I made it more clear on what is returned. So now it has method `platform.getPreferredPlatform()`. I don't think API (even such internal) should expose such implementation details as folder.

If you take a look at folders array, I added new element:

~~~JS
var folders = {
  'osx': 'osx',
  'darwin': 'osx',
  'linux': 'linux',
  'sunos': 'sunos'
};
~~~

So now it is safe to run as much as you can
~~~JS
var platform = folders[getPlatformFromEitherConfigOrOS()];
platform=folders[platform];
platform=folders[platform];
plaform=folders[platform];
etc.
~~~

igorshubovych:
I merged it. We can refactor later if you want.

PR #31, opened 2015-12-21
Move static site url from code to configuration

igorshubovych:
Currently URL to static site is hardcoded.

It is OK for now, but moving it to configuration gives the following benefits:

- redeclare this URL in `$HOME/.tldrrc` file
- allow to move away from static site and migrate to AWS S3, CDN or other type of distribution
- allow to use several pages archives at the moment and easily switch between them, i.e. tldr-latest.zip, tldr-stable.zip

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,5 @@
 {
+  "url": "http://tldr-pages.github.io/assets/tldr.zip",
   "colors": {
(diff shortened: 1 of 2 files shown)


rprieto:
Good idea @igorshubovych :+1:

PR #32, opened 2015-12-21
Adding ESLint, basic linting of existing code base

igorshubovych:
- [x] Added the latest ESLint
- [x] Added recommended configuration for ESLint
- [x] Added `npm run lint` script to simplify linting (I will later add it to pre commit hook)
- [x] Cleaned up the existing code base

Racoon

### DIFF
--- a/bin/tldr
+++ b/bin/tldr
@@ -46,14 +46,21 @@ program.on('--help', function() {
 
 program.parse(process.argv);
 
-     if (program.list)              tldr.list();
-else if (program.random)            tldr.random();
-else if (program.randomExample)     tldr.randomExample();
-else if (program.clearCache)        tldr.clearCache();
-else if (program.update)            tldr.updateCache();
-else if (program.render)            tldr.render(program.render);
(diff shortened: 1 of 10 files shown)


bin/tldr
```diff
@@ -46,14 +46,21 @@ program.on('--help', function() {
...
 
-     if (program.list)              tldr.list();
-else if (program.random)            tldr.random();
-else if (program.randomExample)     tldr.randomExample();
-else if (program.clearCache)        tldr.clearCache();
-else if (program.update)            tldr.updateCache();
-else if (program.render)            tldr.render(program.render);
-else if (program.args.length == 1)  tldr.get(program.args[0], program)
-else {
+if (program.list) {
```
rprieto: I actually like the former style better in this particular case, but :+1: for making the linter happy.

rprieto:
Great idea!

bin/tldr
igorshubovych: If you want I can wrap this code with eslint-ignore.

PR #33, opened 2015-12-22
Fix issue with 'page not found message'

igorshubovych:
Currently if you run 
`$tldr qwerty`

It shows the following message:
```
Page not found
Try updating with "tldr --update", or submit a pull request to:
http://github.com/undefined
```

This PR:
- introduces pagesRepository config option
- points it to tldr-pages/tldr Github repo
- uses it in lib/message.js

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,5 @@
 {
+  "pagesRepository": "tldr-pages/tldr",
   "colors": {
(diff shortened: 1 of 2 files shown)


rprieto:
Good fix, nice one :+1:

Issue #44, opened 2015-12-30
Option to output everything


amitzur:
First, this is rad :)
Since every command is byte size, I wanted an option to output everything and read it out to find out about commands I didn't know.
I tried to add an '--all' option, but then realized that cache.list returns also unavailable commands (e.g. apt-get on osx).
I think cache.list should return only platform-relevant commands. Then iterating over cache.list with tldr.get is a trivial implementation of this feature.
If you think this is a good idea I'll submit a pull request

igorshubovych:
Good idea. Thank you.

amitzur:
Should `trdr --list` return only the platform-relevant commands? Then it's another issue

igorshubovych:
Closing this as the PR is closed.

PR #45, opened 2015-12-30
re #44 add option to output all data

amitzur:
see issue #44 
I thought I would run exports.get on every result from cache.list, but then I resorted to just `printBestPage` even if it's not in the platform.

Now you can run `tldr --all > all` to have a file with all data or `tldr --all | less` to see it in the terminal. Colors are lost, though :(

### DIFF
--- a/bin/tldr
+++ b/bin/tldr
@@ -16,6 +16,7 @@ program
   .option('-e, --random-example', 'Show a random example')
   .option('-f, --render [file]', 'Render a specific markdown [file]')
   .option('-o, --os [type]', 'Override the operating system [linux, osx, sunos]')
+  .option('-a, --all', 'Show tl;dr for all commands in the cache')
   //
   // CACHE MANAGEMENT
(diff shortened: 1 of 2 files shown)


igorshubovych:
After seeing a commit I doubt it is a good idea to merge it.

IMO it is an esoteric option which will be used by really a few people or even you only.

I have an idea how to solve your use case without adding such an option:
- we fix `--list` to print only platform commands
- we add `--list-all` to print all commands
- you are writing simple script (i can help you with it), which use `tldr --list-all`, then makes it single-line output with `tr`, and then with `xargs` and `for` do `tldr $command`.

igorshubovych:
@amitzur what do you think about it?

amitzur:
I think you're right. Achieving what I wanted with a script makes more sense. The more interesting part is to show only platform commands on --list

amitzur:
still I think we should fix #50

PR #47, opened 2015-12-30
Allow searching for multi-word commands

mcm69:
e.g. `tldr git commit`

### DIFF
--- a/bin/tldr
+++ b/bin/tldr
@@ -63,8 +63,8 @@ if (program.list) {
   tldr.updateCache();
 } else if (program.render) {
   tldr.render(program.render);
-} else if (program.args.length == 1) {
-  tldr.get(program.args[0], program);
+} else if (program.args.length >= 1) {
+  tldr.get(program.args.join('-'), program);
 } else {
   util.log('Usage: tldr <command>');
   process.exit(1);


berkus:
Linked https://github.com/tldr-pages/tldr-cpp-client/pull/5

waldyrious:
+:100: this would be very convenient indeed!

Leandros:
All integrated. Update published via homebrew.

rprieto:
I like the idea. It's a tricky one though, given that some complex commands have a page per sub-command (git-log, git-push) but some have a single page with everything (docker). Maybe joining is better than not doing anything though :)

amitzur:
:-1: man doesn't support it, and I think the commit is specific to git. Isn't it the case that not all multi word commands should be joined with a dash?

mcm69:
@amitzur well, this is not `man` anyway. Also, as we do have pages for `git commit` and many more multi-word commands (which are saved in the format of `git-commit.md`), it did not make much sense searching for them as `tldr git-commit`. There's no such command as `git-commit` after all.

igorshubovych:
Thank you for contribution. It gave me an idea on to do it.
There is more sophisticated checks done in #80.

PR #48, opened 2015-12-30
Update package.json

shybovycha:
Forcing `update` right after installing package. This reduces one install step.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   "scripts": {
     "precommit": "npm run lint && npm run test:quiet",
     "prepush": "npm run test:quiet",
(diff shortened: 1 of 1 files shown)


waldyrious:
:+1

mguida22:
This would be a much nicer experience!

amitzur:
:+1:

igorshubovych:
Thanks! It is almost good to go.

But can you please remove `NODE_ENV=development`, because it fails on Windows.

shybovycha:
@igorshubovych Done. Not sure if that env variable should be present in other script commands, so I left them untouched.

igorshubovych:
Thanks for your contribution!

PR #54, opened 2016-01-01
Don't highlight random options

igorshubovych:
`--random`, `--random-example` options are not the most usable commands.

I removed them from README and help text. They are still listed on help page and works. We just not push users to use them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,6 @@ To see tldr pages:
 
 - `tldr <command>` show examples for this command
 - `tldr --list` show all available pages
-- `tldr --random` show a page at random
-- `tldr --random-example` show a single random example
 
 The client caches a copy of all pages locally, in `~/.tldr`.
 There are more commands to control the local cache:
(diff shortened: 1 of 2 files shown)


waldyrious:
Not really a fan of undocumented features. What do we gain by hiding them from these places?

rprieto:
I've actually seen feedback on Twitter that they were fun commands to use (e.g. command of the day / "did you know")

igorshubovych:
I do not remove them completely.

Before
http://take.ms/FYSR1

After:
http://take.ms/Ms7XM

They are listed among the options, just not highlighted.

waldyrious:
Yes, that was clear from your original comment. I still think we shouldn't remove them from the README unless we have good reasons to do so.

igorshubovych:
This is a good reason - documenting actually useful options (#55).

Also there are a lot of questions on color themes, Windows support. This all would be good to document too.

waldyrious:
Well, my stance is that the documentation isn't crowded enough to warrant removal of options at the moment, so I vote to keep them in.

agnivade:
Agree with @waldyrious on this one. Agree that work needs to be done on themes. I will take that up and smooth out any rough edges.

PR #65, opened 2016-01-26
List options

igorshubovych:
New and updated options:
- `-l` or `--list` - print all commands for default platform
- `-a`, `--list-all` - print all commands for all platforms
- `-1` or `--single-column` - print commands in a single column, has to be used together with `--list` or `--list-all`


igorshubovych:
I rebased too much.

Issue #50, opened 2015-12-31
tldr --list should return only platform-relevant commands


amitzur:
Either that or provide an indication for commands that are platform-specific.
I tried to run svcs on my OSX until I realized it's in the sunos folder :)

igorshubovych:
@amitzur do you still consider this a problem?

amitzur:
Yes as discussed in pull request #45

igorshubovych:
In your opinion, what should be the command to print all the commands?

amitzur:
I think you were right suggesting --list-all

igorshubovych:
Upgrade to 1.5.1

PR #66, opened 2016-01-26
--list, --list-all and --single-column options

igorshubovych:
New and updated options:
- `-l` or `--list` - print all commands for default platform
- `-a`, `--list-all` - print all commands for all platforms
- `-1` or `--single-column` - print commands in a single column, has to be used together with `--list` or `--list-all`

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -3,14 +3,15 @@ var path      = require('path');
 var rmdir     = require('rimraf');
 var cpr       = require('cpr');
 var wrench    = require('wrench');
+var _         = require('lodash');
 var config    = require('./config');
 var remote    = require('./remote');
+var platform  = require('./platform');
 
 var CACHE_FOLDER = path.join(config.get().cache, 'cache');
 
 // It's OK to read it on startup
 exports.lastUpdate = lastUpdate();
-
 
 exports.get = function(potentialPages, done) {
   tryLoad(potentialPages, 0, function(err, contents) {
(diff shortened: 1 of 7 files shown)


igorshubovych:
It fixes #50.

igorshubovych:
Sorry for useless commits, i tried really hard (#65).

lib/cache.js
```diff
@@ -62,9 +72,14 @@ function tryLoad(pages, index, done) {
 }
 
 function isPage(file) {
-  return file.match(/pages\/.*\.md$/);
+  return file.match(/.*\.md$/);
 }
 
+function onPlatform(os) {
+  return function(file) {
+    return _.startsWith(file, 'common') || _.startsWith(file, os);
```
rubenvereecken: Tssk including the whole of lodash for this :P? 

`file.match(/^common/)` for the first bit, second you can build from string.

I really don't care though, lodash is great.

rubenvereecken:
Great stuff :). I managed to (kinda) duplicate the `-1` using `tldrl -l | tr ',' '\n'` because I was curious if it was needed, but great PR.

igorshubovych:
Yes, I know that `tr` can help here. But it is so convenient.

rubenvereecken:
I merged it because it was so pretty.

lib/cache.js
igorshubovych: Lodash is used in different places already.

PR #68, opened 2016-01-28
Performance improvement

igorshubovych:
- Load NPM modules only when you need it. 
- Replace string.js with lodash.js, which is already used.

Approximate improvement for standard viewing page is 5%.


rubenvereecken:
5% faster? I wonder what else could be optimized then because it's never been that fast.

:+1: for lazy loading.

rprieto:
I'm not sure about lazy loading of `npm` modules... seems it would make finding which modules are used a lot harder. For a 5% gain I don't know if it's worth it?

igorshubovych:
Seems like really not worth doing.

PR #69, opened 2016-01-31
Multi-words commands

igorshubovych:
It show the correct page when you type
```bash
$ tldr git add
```

You may wonder why I created such a complicated solution instead of just few lines in #47. There are reasons.

- It builds index
- The index is different from classic index.json, it is shorter, faster to load, faster to work with (object vs array)
- The index can be used in future for other commands, i.e. `--list`, maybe in future `--has-page`


igorshubovych:
I decided to redo it and split into 2 parts: index and the rest.

PR #71, opened 2016-02-01
Platforms shortcuts

igorshubovych:
3 command-line options shortcuts for operating systems:
- `--linux` == `--os=linux`
- `--osx` == `--os=osx`
- `--sunos` == `--os=sunos`

So now you can run:
`$ tldr du --linux`
`$ tldr du --osx`


waldyrious:
:+1: nice feature :)

PR #75, opened 2016-02-04
Libraries cleanup and update

igorshubovych:
- remove `async` module
- remove `strings` module, use `lodash` instead
- migrate to modularized `lodash`
- remove `mkdirp` module, use `wrench` instead
- remove `cpr` module, use `wrench` instead
- update `request`, `mocha` and `should`

Tested on all NodeJS versions on Linux, latest Node on OSX, latest Node on Windows 8.1 (with Git Shell).

P.S. Please do not merge it yet.


rubenvereecken:
Every lodash function separate? Sounds hellish. You're brave to try. Looks good to me.

igorshubovych:
The full lodash is overkill.
Another possibility - custom builds.

PR #76, opened 2016-02-04
doc: update usage with the recent commands

igorshubovych:
Added examples of `--linux`, `--osx`, `--sunos` options.

@rubenvereecken, @waldyrious could you, guys, please take a look? I am not very good in English.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,10 @@ $ npm install -g tldr
 To see tldr pages:
 
 - `tldr <command>` show examples for this command
-- `tldr <command> --os=linux` show page for Linux version of this command
+- `tldr <command> --os=<platform>` show command page for given platform (`linux`, `osx`, `sunos`)
+- `tldr --linux <command>` show command page for Linux
+- `tldr  --osx <command>` show command page for OSX
+- `tldr --sunos <command>` show command page for SunOS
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,7 +56,8 @@ This file has to be valid JSON:
 }
 ```
 
-If you need to always read pages for different platform (e.g. Linux), you can put it in config file:
+If you need to always read pages for different platform (e.g. Linux),
+you can put it in config file:
```
waldyrious: "for *a* different platform", "in *the* config file"

README.md
```diff
@@ -21,7 +21,10 @@ $ npm install -g tldr
 To see tldr pages:
 
 - `tldr <command>` show examples for this command
-- `tldr <command> --os=linux` show page for Linux version of this command
+- `tldr <command> --os=<platform>` show command page for given platform (`linux`, `osx`, `sunos`)
```
waldyrious: "for *the* given platform"

waldyrious:
LGTM, only minor comments added inline.

README.md
rubenvereecken: While we're at it.. "If you regularly need pages for a different platform" or something like that, "to always read" won't work.

PR #82, opened 2016-02-09
Minor libraries update

igorshubovych:


rubenvereecken:
Whatever. Use vaguer version specs?

igorshubovych:
Vaguer version specs can cause issues on some combinations of NodeJS/platform.
Remember that we build it on major Node versions and on different systems.

PR #84, opened 2016-02-15
Migration to ESLint 2.0, stricter linting rules and refactoring

igorshubovych:
- Migration to ESLint 2.0
- stricter linting rules
- refactoring according to this rules


ostera:
Looking good!

PR #96, opened 2016-04-01
Ignoring useless files during npm install

igorshubovych:
https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,8 @@
+.github/
+.idea/
+test/
+.editorconfig
+.eslintrc.json
+.travis.yml
+appveyor.yml
+screenshot.png


waldyrious:
@zdroid any thoughts on this one?

zdroid:
@waldyrious You can do it on 2 ways: by adding `.npmignore` or by adding `files` array to `package.json`. The files array would look like this:
```json
"files": [
  "autocompletion",
  "bin",
  "config.json",
  "lib",
  "LICENSE.md"
] 
```

Want me to submit a PR for that?

waldyrious:
If you believe that makes more sense (it seems neater to me, one less file to keep around), yes, please do :)

zdroid:
There you go.

agnivade:
Hmm .. I would have used the .gitignore/.npmignore file for this.

My logic is a blacklist is easier to maintain in this case rather than a whitelist. In future, when we are adding/removing more folders, this place needs to be updated also.

zdroid:
Actually not at all @agnivade. The main directories are lib/ and bin/ and they'll always be included, besides the important metafiles. Basically, 90% of the file additions/deletions will be done in the already-included files.

agnivade:
Well, lets just say that I partially agree. :) Just that using .gitignore/.npmignore seemed a cleaner solution to me. And I have no strong objections against this solution anyways.

PR #100, opened 2016-10-01
Fix indentation in .eslintrc.json

zdroid:
Standard tab size should be 2 spaces, not 4.


waldyrious:
Thanks, much neater 👍

PR #103, opened 2016-10-02
Fix the README to a better way of installation

agnivade:
Asking the user to pass `--ignore-scripts` and then doing `tldr --update` does not look very elegant. A better way is to pass the `--user` variable which will let npm use that user permission while executing the postinstall script. 

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ npm install -g tldr
 If you are installing with sudo, pass the `--user` parameter:
 
 ```bash
-sudo npm install -g tldr --user=<logged_in_user>
+sudo npm install -g tldr --user=$(whoami)
 ```
 
 This is required because we populate the page cache after installation. And npm by default downgrades the post-install user permission level to `nobody` when run as root.
--- a/README.md
+++ b/README.md
@@ -17,12 +17,13 @@ A `Node.js` based command-line client for [tldr](https://github.com/tldr-pages/t
 npm install -g tldr
 ```
 
-If you have trouble with the post-install script, try the following commands:
+If you are installing with sudo, pass the `--user` parameter:
 
 ```bash
-npm install -g --ignore-scripts tldr
-tldr --update
+sudo npm install -g tldr --user=<logged_in_user>
 ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,13 +17,14 @@ A `Node.js` based command-line client for [tldr](https://github.com/tldr-pages/t
 npm install -g tldr
 ```
 
-If you have trouble with the post-install script, try the following commands:
+If you are installing with sudo, pass the `--user` parameter:
 
 ```bash
-npm install -g --ignore-scripts tldr
-tldr --update
+sudo npm install -g tldr --user=<logged_in_user>
```
waldyrious: how about using `whoami` so this command can be used directly?

zdroid: `sudo npm install -g tldr --user=$(whoami)` should do the trick.

agnivade: Yea I had thought about it. But I think it won't work in windows.

zdroid: @agnivade Isn't that an irony? tldr isn't meant to be used on Windows. :smile:

agnivade: Well, that shouldn't stop users from learning about *nix commands from Windows. :)

waldyrious: You mean in `cmd.exe`? I assumed more advanced emulators (cmder/ConEmu, cygwin) support this -- am I mistaken?

agnivade: Yes I meant `cmd.exe`.

waldyrious: I don't think we should be aiming for support in `cmd.exe` -- does the rest of the tool even work there (e.g. the color-coded output)?

agnivade: Yes, color-coded output works. Both in cmd and powershell. Just checked on friend's laptop. But surprisingly `whoami` also works.

But wait, the story doesn't end there. :laughing: Actually, windows doesn't even need the `sudo` part. Simply `npm install -g` works fine without any issues.

So I am guessing its okay to use $(whoami).

waldyrious: Haha wow. I wonder if they're doing something behind the scenes? Anyway, great find :)

PR #118, opened 2016-11-05
Build the snap in travis

elopio:


elopio:
Building it is the first step to continuously deliver the snap to the store. But to actually publish it you will need to register the name: https://myapps.developer.ubuntu.com/dev/click-apps/register-name/

Are you still interested in publishing it?

agnivade:
I am so sorry man .. totally forgot about this ! 

Yes, I am interested in publishing it. I will register the name.

elopio:
Thanks @agnivade. I would suggest you to publish one snap to the edge channel, so you get an idea of how it works. Then we can hook travis to land there after every change.

agnivade:
Hey @elopio , I registered an account. Now I am trying to build a snap. It gives me -
`Issues while validating snapcraft.yaml: 'vendor' is a required property`

Can you help me out here ? I am using Ubuntu 15.10. Google search tells me I need to use the 1.x version of snapcraft. It not very clear how I am supposed to do that.

elopio:
Thanks for taking the time to go into the store.
We are no longer maintaining snapcraft 1. You should use snacraft 2, which is in Ubuntu 16.04 and later. I would recommend to upgrade to 16.04, because 15.10 has reached it's end-of-life in july, so you are not getting any security updates there.
If you can't or don't want to upgrade, you can also use the same trick we are doing here in travis, which is to use a docker machine to get an ubuntu container with snapcraft 2.

If you have some time, would you like to jump into a hangout or a chat? We can get this done, talk some more about the daily releases to the edge channel, and answer any question you have.

pura vida

agnivade:
Thanks, I will use the docker approach. Btw, is there a reason snapcraft 2 is not in Ubuntu 15.10 ?

elopio:
@agnivade the reason is that we started the snapcraft 2 branch along the development cycle of 16.04, which is the long-term release. We introduced some dependencies into the distro just for snapcraft during that cycle.

The currently active Ubuntu versions are 14.04, 16.04 and 16.10. snapcraft is not yet on 14.04, but we are working on some changes that will make it installable there.

agnivade:
@elopio - Created and published the snap here - https://myapps.developer.ubuntu.com/dev/click-apps/6419/. One thing I noticed was that there was no MIT license in the license dropdown. You guys might want to add that.

Anyways, let us go ahead with the publishing.

elopio:
Nice @agnivade! 

I see the mit license there:
selection_002
Maybe I'm looking at the wrong place? Can you please check.

Now that you registered, there are two options for building the snap. You can use travis, which will push only for the amd64 platform but it's a good start, or you can use launchpad, which will build for all the architectures. This is kind of a work in progress, so your feedback will be really useful for us. Why don't you jump into https://rocket.ubuntu.com/channel/snapcraft, and ping me when you have the time to talk about it?

agnivade:
> I see the mit license there:

Ah yes, but your options are not sorted alphabetically ! :wink: 

Let us start with building the snap in travis and pushing only for amd64. I can see that your PR only builds it. Can you update it so that it pushes it also ? Lets get this to working state, and then I will look into launchpad.

elopio:
@agnivade ah, that's a good point. It seems it tries to show first the most common ones, which in the end result in confusion. I will talk with the store guys.

I can't make the push PR, because it needs your credentials. We have a new unreleased snapcraft command called enable-ci. Would you like to give it a try?
However, if you are still in 15.10, you will need at least a lxc container of 16.04, which is not hard to get.

As of this branch, it can land, and then we add on top of it the push logic.

agnivade:
@elopio - It seems that you guys are still in the process of fine tuning the process. There seems to be different tools supported in different platforms which is making this process a bit difficult for me. I do not want to test this whole 15.10-16.04 thing. All pushes will be done through travis for now. For that, you already have a 16.04 environment.

Please add the push commands, with the required login details being taken from environment variables. I will set them from travis settings.

elopio:
Yes, we are very much in active development to simplify the integration with CI systems.
I added the push on travis. The thing I can't do for you is running the seed script. And that needs 16.04 or later so you get the valid credentials. We can't do anything to help you on 15.10, because that version has been closed for four months already.

agnivade:
Thanks, that's what I needed.

waldyrious:
@agnivade I'm not sure if this is relevant, but you might want to take a look at https://github.com/FredrikNoren/ungit/issues/823 (and the PR implementing it, https://github.com/FredrikNoren/ungit/pull/827)

elopio:
@agnivade I've just came back from holidays, and I'm checking on the pending discussions from last year.
Looking at the store I see you have pushed one revision, but haven't published it. Could you publish it to start testing it here in a few devices? I'll be available again in case you need any help.

Happy new year!

elopio:
I'm closing this, because it's no longer necessary to do the push to the store from travis.
Now we have https://build.snapcraft.io, that will take care of building and pushing every change in master, with a couple of clicks. Give it a try! :)

PR #121, opened 2016-11-25
Newformat

agnivade:


test/parser.spec.js
```diff
@@ -41,9 +41,9 @@ describe('Parser', function() {
       '\n# tar' +
       '\n> archiving utility' +
       '\n' +
-      '\n- create an archive' +
+      '\ncreate an archive' +
```
waldyrious: Shouldn't the first letter be capitalized?

agnivade: I think that's the work of the linter. The client is not bothered about it. It will display verbatim.

waldyrious: Sure, I understand that it doesn't care, but since this code is meant to represent real tldr-pages content, it would make sense IMO to match the actual format, for consistency. But of course, that change is not directly related to this PR, so it shouldn't hold up merging.

agnivade:
This is unlikely to happen in the near future. Let's just close it and we can revisit this if we start to work on it again.

PR #122, opened 2016-12-28
Removing postinstall step

### LINKED ISSUE #119: Possible to do the initial --update on first run instead of on postinstall?
Instead of running the initial `--update` on postinstall, I wonder if it's possible to do so on first run instead? It would cause the least amount of surprise that way.

Right now, installing this package with sudo requires running a different command than `npm install -g tldr`. This is not mentioned in [tldr's main site](http://tldr-pages.github.io), and breaks certain flows (e.g. mass-install npm packages via devops tools).

agnivade:
- This was a source of errors for a long time. Now, we update the cache
on a page fetch miss.

Fixes #119 and #9

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,17 +14,8 @@ A `Node.js` based command-line client for [tldr](https://github.com/tldr-pages/t
 ## Installing
 
 ```bash
-npm install -g tldr
+sudo npm install -g tldr
 ```
-
-If you are installing with sudo, pass the `--user` parameter:
-
-```bash
-sudo npm install -g tldr --user=$(whoami)
-```
-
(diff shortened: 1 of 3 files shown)


waldyrious:
Nice! +:100:

Please also fix the install instructions on the README.

waldyrious:
I don't have any concerns, but others more versed in node.js may want to chime in. Either ping someone for further review,  or merge if you feel confident on these changes :)

agnivade:
ping  @tldr-pages/node-client for a second look.

waldyrious:
Wooo @igorshubovych! Are you back or just passing by?

PR #125, opened 2017-01-04
WIP: Codebase revamp and moving to ES6

### LINKED ISSUE #110: Update codebase to ES6 standard
Node is EOLing 0.10 on Oct end. This is a good opportunity to make the minimum version to be 4. And make the codebase to be readable and maintainable.

agnivade:
Fixes #110. Commits will be added as I update each module one by one.

General changes are:
- moving to async commands.
- updating code to reflect ES6 syntax.

Note - I haven't used the `import` syntax yet because I am supporting 4.7 as the min version.


agnivade:
@zdroid - Finally the time has come. This is still a work in progress. But you can take a look if you want.

zdroid:
Looks mostly fine to me. Just one thing to fix: use `(arg) => {}` syntax for anonymous functions. It saves a lot of space, gets rid of bad coding style (where `function` isn't separated from parens) and makes code fancier and more readable. :wink:

agnivade:
Yes, been doing that mostly. Sometimes, one or two fall through. Will do a thorough check when I am done with the other pages.

zdroid:
Awesome progress so far. If I notice something that hasn't been ES6-ified, I'll post it here.

lib/config.js
```diff
@@ -1,44 +1,51 @@
-var defaults = require('lodash.defaults');
-var compact = require('lodash.compact');
-var map = require('lodash.map');
-var fs        = require('fs');
-var path      = require('path');
-var osHomedir = require('os-homedir');
+'use strict';
 
-var config = null;
+const defaults = require('lodash.defaults');
```
waldyrious: we'll eventually want to align this equals sign with the ones below, right?

agnivade: Not unless that is actually forced by the linter. For example, the golang's linter(gfmt) automatically does this. For now, this is not the priority.

waldyrious: Sure, not a priority, just wondering since the ones below are aligned. You were planning to do such formatting cleanups after the main conversion was complete (or would agree if they were made after this PR is merged. Just to be clear: would you be opposed to a code style cleanup PR afterwards?

agnivade: Why would I be opposed ? :laughing: Unless something automatically does this, I would actually unalign them since that's the natural way of writing. Will take care of it in a later PR.

waldyrious: To clarify, I meant code style cleanup including alignment of operators like this. It's pretty common to manually align assignments, function parameters, etc., for readability.

Responding more directly to "Why would I be opposed?", the question was due to this exchange:

> -- We'll eventually want to align this equals sign with the ones below, right?
> -- Not unless that is actually forced by the linter.

I read that as "we will not want to align this equals sign with the ones below", which I found odd (hence my question), but "I would actually unalign them" confirms that I read you correctly. In any case, this is a rather minor issue so I wouldn't insist on any particular format. Carry on as you prefer :)

PR #129, opened 2017-01-16
Add --markdown output option

twang2218:
## Description

Sometimes, we need the raw markdown output, such as another library consume the `tldr`'s output. For example, the [tldr plugin for Zazu](https://github.com/twang2218/zazu-tldr) need to have parsable output for processing.

So be able to have a raw markdown output would be handy.

I add `--markdown` or `-m` for short, for the raw markdown output option.

## Checklist

- [x] Code compiles correctly

Signed-off-by: Tao Wang <twang2218@gmail.com>

### DIFF
--- a/bin/tldr
+++ b/bin/tldr
@@ -19,6 +19,7 @@ program
   .option('-r, --random', 'Show a random command')
   .option('-e, --random-example', 'Show a random example')
   .option('-f, --render [file]', 'Render a specific markdown [file]')
+  .option('-m, --markdown', 'Output in markdown format')
   .option('-o, --os [type]', 'Override the operating system [linux, osx, sunos]')
   .option('--linux', 'Override the operating system with Linux')
   .option('--osx', 'Override the operating system with OSX')
--- a/lib/tldr.js
+++ b/lib/tldr.js
@@ -24,8 +24,8 @@ exports.listAll = (singleColumn) => {
   });
 };
 
-exports.get = (commands) => {
-  printBestPage(commands.join('-'));
+exports.get = (commands, options) => {
+  printBestPage(commands.join('-'), options);
 };
 
 exports.random = () => {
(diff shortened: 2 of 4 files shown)


waldyrious:
Thanks for the contribution! @agnivade will likely have the final word on this, but personally I'd prefer if such an option was called `--raw`.

twang2218:
Originally it is `--raw`, but I'm not sure the `raw` is clear enough to indicate that it's actually a markdown format out, so I replace it with `--markdown`. But both works for me.

waldyrious:
Sure. Let's wait and see what @agnivade thinks.

bin/tldr
```diff
@@ -92,7 +93,11 @@ if (program.list) {
 } else if (program.render) {
   tldr.render(program.render);
 } else if (program.args.length >= 1) {
-  tldr.get(program.args, program);
+  if (program.markdown) {
```
agnivade: No need to specify an if condition here. We are already passing the `program` variable to `tldr.get` which contains the markdown option. Let us do the final calculation in the `renderContent` function.

agnivade:
Please leave out the other changes. Make the changes in `tldr.js` that I have mentioned. Of course adding the `--markdown` option in tldr file and the autocompletion file will still be there. Rest can be left out.

twang2218:
@agnivade Yes, it is much cleaner in your implementation, I changed the code by using `options` arguments, and removed the accidentally committed `package.json` file.

Could you have a look, is there anything else should I change?

lib/tldr.js
```diff
@@ -24,8 +24,8 @@ exports.listAll = (singleColumn) => {
   });
 };
 
-exports.get = (commands) => {
-  printBestPage(commands.join('-'));
+exports.get = (commands, options) => {
+  printBestPage(commands.join('-'), options);
```
agnivade: Please add this line inside printBestPage function to convert to empty dictionary if `options` is not passed.

```
options = options || {};
```

to fix the travis error.

twang2218: Ok, it's done. And Travis CI is happy now. 😺

PR #142, opened 2017-03-14
Remove sudo from installation method

ran-y:
## Description
removed the sudo from the readme installation area.

using sudo to install npm packages is not a best practice, see:

* https://docs.npmjs.com/getting-started/fixing-npm-permissions
* https://github.com/sindresorhus/guides/blob/master/npm-global-without-sudo.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,8 +134,6 @@ fpath = (my/completions $fpath)
 
 You probably have a permission problem, which you can solve [Here](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
 
-You can also use `sudo npm install -g tldr` to install, but [this pull request](https://github.com/tldr-pages/tldr-node-client/pull/142) will show you why you shoudn't.
-
 #### Colors under Cygwin
 
 Colors can't be shown under Mintty or PuTTY, because the dependency `colors.js` has a bug.
--- a/README.md
+++ b/README.md
@@ -130,6 +130,12 @@ fpath = (my/completions $fpath)
 
 ## FAQ
 
+#### `npm install -g tldr` throws an error
+
+You probably have a permission problem, which you can solve [Here](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
+
+You can also use `sudo npm install -g tldr` to install, but [this pull request](https://github.com/tldr-pages/tldr-node-client/pull/142) will show you why you shoudn't.
+
 #### Colors under Cygwin
 
 Colors can't be shown under Mintty or PuTTY, because the dependency `colors.js` has a bug.
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ A `Node.js` based command-line client for [tldr](https://github.com/tldr-pages/t
 ## Installing
 
 ```bash
-sudo npm install -g tldr
+npm install -g tldr
 ```
 ## Usage
 


agnivade:
Hi @ran-y, yes I am aware of that. However, majority of users of npm out there, install global packages with `sudo`. So, for the sake of simplicity, I have decided to keep it.

waldyrious:
Particularly, I'd point out guideline n.2 of the tldr-pages project, which says:

> When in doubt, keep new command-line users in mind. Err on the side of clarity rather than terseness. For example, commands that require sudo should include it directly in the examples.

The sudo example is not the same as this, but the principle behind it is similar.

ran-y:
Hey @agnivade and @waldyrious, thank you for the quick response.

First, I don't know if the majority install global packages with sudo.
I know it's not a good thing, for a variety of reasons, please have a look at:

https://givan.se/do-not-sudo-npm/
http://hood.ie/blog/why-you-shouldnt-use-sudo-with-npm.html
https://www.gregjs.com/javascript/2016/friends-don-t-let-friends-sudo-npm-install-g-packages/

Which explain why it's a bad practice.
About guideline n.2, I think we should remove the sudo because of it and not despite it.
New users should be instructed by best practices.

when you perform `sudo npm install -g`, any future global installations with npm install -g will need root permissions.

You should not run applications with root privileges if its not necessary. Node and npm can do their work just perfectly fine without admin powers.

waldyrious:
@ran-y but won't installing without `-g` (which requires sudo, IIRC) make the tldr not accessible system-wide? It seems to me that is the expected behavior when a user installs a program, so by recommending otherwise we'd be unnecessarily confusing them. How about using sudo's -H option?

ran-y:
@waldyrious, of course the -g flag is required, no arguments on that.
I was just suggesting to remove the `sudo` prefix.

ostera:
Actually guys, I'm afraid that this is one of the cases where rooting for the newbie is only going to encourage and reinforce lousy behavior.

(Okay maybe _lousy_ is a bit too much, but you get the point)

So I'm going to turn it around: can we measure this? And if so, would it be okay to experiment with this and see how it impacts the install numbers / issues with installing the client?

waldyrious:
I'm just trying to understand what's the alternative here. If we keep `-g`, won't not using `sudo` produce an error for most people? What am I missing?

agnivade:
@waldyrious - Yes, if we keep `-g` and `sudo` both, for most people it will cause an error. And that is @ran-y's point. He wants new users to be directed to a best practice by sacrificing usability. 

@ostera - Good point. I did a cursory look on top cli packages on npmjs, and almost none of them use `sudo`. And also our tldr github page too doesn't use `sudo` for the tldr npm package install step.

I think it makes sense to remove `sudo` to be uniform with other packages out there, and also maintain uniformity internally.

ran-y:
@ostera @waldyrious @agnivade 

The users can be divided into three groups:
1. never installed any npm pacakge globally.
2. used `sudo npm i -g` once, and now they have to give root privileges for every globally installed package.
3. always use `npm i -g` without sudo.

**For group 1** (the newbies) we will be fully compatible without the sudo.

**For group 3** we will be fully compatible without the sudo.

**For group 2**, this group members or know that something is wrong, because they have to type sudo before any global npm installation (which means they add it themselves as most cli pacakges doesn't use sudo) or they don't, which in this case, we don't need to encourage them to do something we know isn't good for them.

We should give root privilege only when it's truly needed.

waldyrious:
Ok that clarifies things, thanks. I would like if we could somehow point group 2 to how to properly fix the issue, rather than defaulting to the "no permission" error (otherwise we will start getting bug reports or pull requests asking to add the sudo). Do you guys see any way we could provide a custom error in that case?

At the very least, we should add a note above the installation command in the readme linking to one of the resources above.

agnivade:
We can do `[sudo] npm install -g tldr`. I feel its the best of both worlds.

ran-y:
@agnivade - The `[sudo] npm install -g tldr` is better than what we have now, but it implies that this is an ok thing to do.

Another solution will be to add this in the FAQ section or maybe to leave a comment with a relevant information or a link to this discussion.

(which is more than enough, as @agnivade said "almost none of the top cli packages uses `sudo` on the install script" )

agnivade:
That sounds good to me @ran-y. Can you update this PR adding a line about that in the FAQ section ?

waldyrious:
> adding a line about that in the FAQ section ?

I'd prefer if the note was immediately above the installation command in the readme, with a link to one of the pages @ran-y mentioned above. It could be included in the FAQ as well, of course.

ran-y:
@waldyrious 

I tried to add a link above the installation section, but it just looked a bit unprofessional.
I think the FAQ section should be enough.

If you have a neat way to add it, please do.

waldyrious:
I don't like the link to this PR, which is a discussion with some back-and-forth, but I'm OK with merging this change and tweaking it later on.

ostera:
I'd say if we include a link to how to fix the `sudo` crap, then don't show how to do the `sudo` crap. Other than that I'm okay with merging and tweaking later too.

agnivade:
@waldyrious , @ostera - Removed the last para which addresses both of your concerns. Lets tweak in a different PR if you want something else too.

PR #150, opened 2017-08-18
index.js: various documentation fixes

waldyrious:
Just made a few minor edits to the comments in index.js, for clarity and consistency.

PR #151, opened 2017-09-09
Fixing failing tests

agnivade:
## Description
Fixed unit tests that were failing
Tied down dev dependencies to minor releases. Keeping it up-to-date is a bad idea and will break code like it did here. We can always intentionally upgrade any package we want.

## Checklist

- [x] Code compiles correctly
~~- [ ] Created tests, if possible~~
~~- [ ] Extended the README / documentation, if necessary~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,11 +63,11 @@
     "unzip2": "~0.2.5"
   },
   "devDependencies": {
-    "env-test": "*",
-    "eslint": "*",
-    "husky": "*",
-    "mocha": "*",
-    "should": "*",
-    "sinon": "*"
+    "env-test": "^1.0.0",
+    "eslint": "^4.6.1",
+    "husky": "^0.14.3",
+    "mocha": "^3.5.0",
+    "should": "^13.0.1",
+    "sinon": "^3.2.1"
   }
 }
--- a/test/cache.spec.js
+++ b/test/cache.spec.js
@@ -42,11 +42,11 @@ describe('Cache', () => {
     });
 
     it('should return page contents for ls', (done) => {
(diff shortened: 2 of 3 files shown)


agnivade:
@waldyrious - Quick review ? You can merge #150 after we push this.

waldyrious:
I'm have pretty much zero node.js development experience, but the changes do look reasonable to me, and tests are indeed fixed. I'll squash-merge to edit the final commit message a bit, hope you don't mind.

waldyrious:
Could we set up greenkeeper to automatically submit PRs to update dependencies when new versions are released?

agnivade:
Ugh .. this is one thing I hate about Node. There's always this cat and mouse game you have to play to keep your dependencies up to date. I'd rather update dependencies when I feel like it. That's why I removed the david-dm dependency checker.

waldyrious:
Yeah but automated PRs are simpler to handle, right?

If tests don't break for those PRs, we can simply merge them and get, say, security updates or bug fixes for edge cases we hadn't been exposed to yet.

If the tests do break, that gives us a heads up and we can start considering how to adjust them (as you did here), rather than being caught by surprise later on.

agnivade:
I understand all that. My point wasn't about simplicity. It's a mental barrier I have that I don't like overdoing software. It doesn't matter that an automated tool is doing it or I am doing it manually, the fact that its been done is what bothers me.

I don't really see any need to always remain up-to-date here. We anyway use a small no. of dependencies. I would prefer if its kept that way.

waldyrious:
Fair enough, I can understand that POV.

Issue #160, opened 2017-11-13
Switch to Promises


aero31aero:
## Expected behaviour
The client should use promises instead of callbacks.

## Actual behaviour
The client uses callbacks predominantly.

## Further details

This is a placeholder issue for the task of switching to Promises. I'd like to suggest using Bluebird Promises for the migration, at least, mainly because of the following function(and its usage):

```js
const Promise = require('bluebird');
const fs = Promise.promisifyAll(require('fs-extra'));
fs.readFileAsync(file).then(data => {
  console.log(data.toString());
}).catch(error => {
  console.error(error);
});
```

This would allow easy and fast migration to promises.

## Progress

- [ ] cache.js
- [x] config.js (not applicable)
- [x] index.js
- [x] messages.js (not applicable)
- [x] parser.js (not applicable)
- [x] platform.js (not applicable)
- [x] remote.js
- [x] render.js (not applicable)
- [x] search.js
- [x] theme.js (not applicable)
- [ ] tldr.js
- [x] utils.js (not applicable)

agnivade:
I am okay with switching to promises to avoid the callback depth. Although its not too much in our case. But please use the native Promise in node. No need to add any new dependency.

vladimyr:
Would you let me to help you with this one? :wink:

Before I dive into technical stuff, let me say few things:

1. I'm longtime (almost on daily basis) happy `tldr-node-client` user, **giant** kudos to all of you who put in your work and effort to make it happen :+1: :100: :trophy:
2. I'm not here to give anyone condescending-ish _programming lessons_ period. I'm sort of jumping straight out of nowhere and doing late _code review_ which is unfortunate by definition but my only motive is giving back to this **wonderful** :heart: project by offering my help. I did write small essay down there but purely to support my case. I hope you'll find it constructive and I honestly hope that by writing this I'm avoiding any unfortunate misinterpretations of my act.
3. I basically rewrote whole article but still it is always good to bring more material into discussion so here you go:
https://hackernoon.com/javascript-promises-best-practices-anti-patterns-b32309f65551
In the same time I want to **strongly** emphasize that what I found inside codebase is light years ahead compared to bad code example from the article.

Back to topic...
There are some recent commits on master branch that bring in required changes. Went through code and noticed particular pattern. Promisification is done by essentially wrapping methods with `new Promise(...)` calls which is fine but those already contained promise chains so now we ended up with nested promise chains, sort of inception thingie :dizzy_face:

[...]

PR #161, opened 2017-11-13
[WIP] Implement Search

aero31aero:
**This is just a work in progress PR to let the community know this is being worked on.**

Implements search and starts migration to promises. Adds the following new dependencies;

1. Bluebird: refer #160
2. Natural: for tokenization, stemming and tfidf calculation.
3. Glob: for easy selection of all cached pages. This could be removed at a later stage with a custom implementation, but for dev, I'll let it stay.

- [x] Code compiles correctly


agnivade:
Thanks for the PR. Please have a look into my [comment](https://github.com/tldr-pages/tldr-node-client/issues/160#issuecomment-343799438)

About the search functionality, is it like the `apropos` command ? i.e. Does it search for page contents also and return the command names whose pages contain those keywords ?

aero31aero:
Yes, like the `apropos` command. Would search within the manual pages and implement basic information retrieval techniques to try to fetch relevant results.

agnivade:
Ok, please use native promises instead of bluebird. And use `let` instead of `var`. 

I will do a review once you add tests and documentation. Thanks.

aero31aero:
Initial implementation now looks complete. Here are the usage instructions:

1. `tldr update`. This would also create the corpus.
2. `tldr search "create symbolic link to file"`.

You need to put the query in quotes if the query contains any defined option as well. For example, `npm start -- search file search` would fail. Looking for a way around that in the current options library.

Still not ready for a review but I'd like to ask what kind of UI would be better. As of now, I just `console.log` the top 10 results as filenames of the cached files and their rank. I had three ideas:

1. Print the top page.
2. List the top 10 results such that typing `tldr <result>` would then get the content of the page.
3. Print the top page and then list the top 10 results.

Feedback appreciated.

agnivade:
Please go with option 2.

Also, how about using an ES6 class for this ? It seems to be particularly well suited for this purpose.

package.json
```diff
@@ -41,7 +41,7 @@
     "example": "node ./bin/tldr tar",
     "test": "mocha test --require=env-test",
     "test:quiet": "npm test --reporter=dot",
-    "lint": "eslint lib test bin/tldr",
+    "lint": "eslint --global Promise lib test bin/tldr",
```
agnivade: Please set this in the `.eslintrc`. Avoid putting settings from the command line.

agnivade: Taken care of in #166. You can remove this.

aero31aero:
Explain what do you mean by using a class here. You mean convert the whole module to a class? I thought that since this file is just one module exporting only two endpoints, it should be fine.

And :+1: for option 2.

package.json
aero31aero: Apologies for this. Didn't notice the `.eslintrc` file earlier.

agnivade:
>  You mean convert the whole module to a class?

Yes. But now that I think of it, you would have to create the object for this in tldr.js and then call the functions. And the object will just remain in global scope. Not really elegant. We can drop the idea. I will revisit it later when we are done with full promisification of the codebase.

aero31aero:
Ready for final review now. I intend to make small incremental improvements overtime. Also, since most of the functions that can be tested are private, I'd have to come up with a testing strategy for them. Perhaps a separate PR for it if this gets merged.

Sample output:

```bash
$ tldr --search create symbolic link to file --os windows
Searching for: create symbolic link to file

    $ ln
    $ update-alternatives (Available on: linux)
    $ nm (Available on: linux, osx)
    $ gcc
    $ chgrp
    $ c99
    $ apg
    $ mkdir
    $ umask (Available on: linux)

Run tldr <command> to see specific pages.
```

agnivade:
We need to have some basic sanity tests atleast. No need to rush through anything. Take your time, come  up with a proper testing strategy. And then we can get to reviewing this.

Thanks again for all your work.

agnivade:
Btw, I have promisified `index.getShortIndex`. So you will need to update your code slightly.

aero31aero:
I was just worried the PR was getting too big and thought would add the tests in another one.

Anyway, pushed the basic integration tests. Have exams soon so would not be able to work on the other tests for some time. If you want to merge only after all the tests are there, I'll write them sometime in December. That was kind of why I was hoping to get this done quick. :P

Also, rebased with latest master and updated for latest changes.

agnivade:
Great. I will take it from here. You can add more tests later.

agnivade:
Closing in favor of #171.

PR #184, opened 2017-11-27
Switch to caret dependencies

billyjanitsch:
## Description

Currently, the dependencies are defined with a mix of caret and tilde ranges. Generally caret dependencies are preferred because they allow the latest non-breaking package versions to be installed.

If you'd prefer to use tilde ranges instead, I can switch them all to use tildes instead. Or is there a particular reason you're using a mix of the two?

## Checklist

- [x] Code compiles correctly


agnivade:
Thanks for taking care of this.

PR #186, opened 2017-11-28
Use modular Lodash imports

### LINKED ISSUE #185: Switch to modular lodash dependency?
Currently, this package depends on several per-method Lodash packages (e.g. `lodash.defaults`).

It's generally better to depend on the main Lodash package and use the modular imports it provides (e.g. `require('lodash/defaults')`), especially if you depend on more than one method. In the upcoming Lodash 5, per-method packages are [being discontinued](https://github.com/lodash/lodash/wiki/Roadmap#v500-2017) entirely.

The reason to prefer modular imports is that the per-method packages are compiled from the Lodash source to zero-dependency files which do not share code with each other. If you look at the [source](https://unpkg.com/lodash.get@4.4.2/index.js) of these packages, they are quite large and each primarily contain the same code setting up Lodash's conveniences like shorthands and iteratees. Modular imports are the best of both worlds, importing [only](https://unpkg.com/lodash@4.17.4/get) the methods you use, but allowing them to share mutually dependent code. They are also simpler to manage, since you don't have to install a new dependency to use a new method.

Would you accept a PR switching to a single dependency on `lodash` and using modular imports?

billyjanitsch:
## Description

Resolves #185.

## Checklist

- [x] Code compiles correctly


billyjanitsch:
@agnivade I rebased this onto latest master, so it should be ready.

PR #187, opened 2017-11-28
Upgrade dependencies

billyjanitsch:
## Description

I was able to bump major versions of all outdated dependencies without breaking anything.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -112,12 +112,6 @@
       "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
       "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU="
     },
-    "async": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/async/-/async-1.0.0.tgz",
-      "integrity": "sha1-+PwEyjoTeErenhZBr5hXjPvWR6k=",
-      "dev": true
-    },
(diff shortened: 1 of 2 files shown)


agnivade:
@billyjanitsch - Can you test with `./bin/tldr -l` and `./bin/tldr -a` and confirm that it works ?

Also, I'm curious whether this fixes the uglify-js dependency error that github is giving after we upgraded to the caret dependencies.

billyjanitsch:
@agnivade Yes, both of those commands work. The only non-dev dependency being updated is `ms`, for which the list of breaking changes is [small](https://github.com/zeit/ms/releases).

Where are you seeing the uglify-js dependency error? I can't reproduce it on master.

agnivade:
> Where are you seeing the uglify-js dependency error?

Hmm .. maybe only the owner can see it.

Can you please rebase the latest master on to this and update the PR ? The ms is only used for a single call. I think I will remove it altogether sometime later.

waldyrious:
>> Where are you seeing the uglify-js dependency error?

> Hmm .. maybe only the owner can see it.

screenshot from 2017-11-29 12-13-58

billyjanitsch:
> Can you please rebase the latest master on to this and update the PR ?

Done! And, yes, this dependency upgrade happens to remove the (transitive) dependency of uglify-js entirely, so the security warning should go away.

PR #198, opened 2018-01-30
Update cache automatically when it's out of date

briandk:
### Current Behavior

If a user runs `tldr [command]` and the cache is stale, tldr prints a warning to the user to run a `--update` command.

Users experience this:

```bash
$ tldr git branch                
Local data is older than two weeks, use --update to update it.
```

because of this:

https://github.com/tldr-pages/tldr-node-client/blob/e7d6519a8eb128d807069ceb534d75283be706de/lib/tldr.js#L195-L205

### Desired Behavior

Instead of telling the user to run a separate command, just automatically try to update the cache and continue as normal.

```js
    if (cacheIsOutOfDate) {
      exports.updateCache();
    };
```

## Checklist

- [x] Code compiles correctly


lib/tldr.js
```diff
@@ -194,12 +194,13 @@ function renderContent(content, options={}) {
 
 function checkStale() {
   cache.lastUpdated((err, stats) => {
+    const cacheIsOutOfDate = stats.mtime < Date.now() - ms('30d');
```
agnivade: Move this below the error check. Rename to `isCacheOutdated`

lib/tldr.js
```diff
@@ -194,12 +194,13 @@ function renderContent(content, options={}) {
...
     if (err) {
       console.error(err);
       exit(1);
     }
-    if (stats.mtime < Date.now() - ms('30d')) {
-      console.warn('Cache is out of date, you should run "tldr --update"');
+    if (cacheIsOutOfDate) {
+      exports.updateCache();
```
agnivade: Use the `cache.update()` call here. It is a bit unidiomatic to use the exported call directly.

lib/tldr.js
```diff
@@ -198,8 +198,9 @@ function checkStale() {
       console.error(err);
       exit(1);
     }
-    if (stats.mtime < Date.now() - ms('30d')) {
-      console.warn('Cache is out of date, you should run "tldr --update"');
+    const isCacheOutdated = stats.mtime < Date.now() - ms('30d');
+    if (isCacheOutdated) {
+      cache.update();
```
agnivade: Please also handle the error in a callback.

```js
    cache.update((err) => {
      if (err) {
        console.error(err);
      }
    });
```

briandk:
I'm happy to make those requested changes. 😃 

One thing I want to point out: *repeated functionality*

If I [handle an error](https://github.com/tldr-pages/tldr-node-client/pull/198#discussion_r165141969), then I feel like the same functionality ends up replicated in three places in the code:

1. [`cache.update()`](https://github.com/tldr-pages/tldr-node-client/blob/e7d6519a8eb128d807069ceb534d75283be706de/lib/cache.js#L37-L55)
2. [`tldr.updateCache()`](https://github.com/tldr-pages/tldr-node-client/blob/e7d6519a8eb128d807069ceb534d75283be706de/lib/tldr.js#L92-L106), which wraps `cache.update()`
3. [Your requested change](https://github.com/tldr-pages/tldr-node-client/pull/198#discussion_r165141969), which also calls `cache.update()` 

What if `tldr.updateCache()` weren't immediately bound to the `exports` object?, so that my code could just call `updateCache()` instead of calling `cache.update()` and needing extra boilerplate to handle errors?

agnivade:
You are saying you will refactor it to another function and then call that ? Sure, that works too.

agnivade:
ping @briandk

briandk:
@agnivade Sorry it took so long, but I finally got around to refactoring the `checkStale()` logic for the cache. 

There's some noise in the whitespace changes, and I'm sorry about that. I run [Prettier][1] pretty aggressively. If you'd like, I can open a PR to add a prettier config file to the repo, so that way formatting should be consistent in the future.

[1]: https://prettier.io

briandk:
This URL should let you view the changes with some of the whitespace change noise damped out:

https://github.com/tldr-pages/tldr-node-client/pull/198/files?w=1

_Note_: I think passing the `w=1` as a URL query string parameter is what helps suppress whitespace changes in the diff view.

agnivade:
Awesome, thanks !

Could you please revert the whitespace changes ? We just use eslint in our repo for formatting. Let's introduce prettier changes in a different PR along with the config file.

Also, please rebase the branch from master.

Thanks.

briandk:
Superseded by #291

PR #199, opened 2018-03-09
Promisify methods without extra wrappers

vladimyr:
## Description
Continuing promisification effort; unwrapping promise chains as described here: https://github.com/tldr-pages/tldr-node-client/issues/160#issuecomment-371617564

Refactored `#readShortPagesIndex` to use `#readJson`/`#writeJson` (from `fs-extra`); makes code even simpler 🏄 

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -9,47 +9,27 @@ const index = require('./index');
 
 const CACHE_FOLDER = path.join(config.get().cache, 'cache');
 
-exports.lastUpdated = (done) => {
-  fs.stat(CACHE_FOLDER, done);
-};
+exports.lastUpdated = () => fs.stat(CACHE_FOLDER);
 
-exports.getPage = (page, done) => {
+exports.getPage = (page) => {
   let preferredPlatform = platform.getPreferredPlatformFolder();
-  index.findPlatform(page, preferredPlatform)
+  return index.findPlatform(page, preferredPlatform)
     .then((folder) => {
-      if (folder) {
-        let filePath = path.join(CACHE_FOLDER, 'pages', folder, page + '.md');
-        fs.readFile(filePath, 'utf8', done);
-      } else {
-        done(null, null);
+      if (!folder) {
+        return;
       }
+      let filePath = path.join(CACHE_FOLDER, 'pages', folder, page + '.md');
+      return fs.readFile(filePath, 'utf8');
     })
-    .catch((err) => {
-      console.error(err);
-      done(null, null);
-    });
+    .catch((err) => console.error(err));
 };
 
-exports.clear = (done) => {
-  fs.remove(CACHE_FOLDER, done);
-};
+exports.clear = () => fs.remove(CACHE_FOLDER);
 
-exports.update = (done) => {
+exports.update = () => {
   // Downloading fresh copy
-  fs.ensureDir(CACHE_FOLDER)
-    .then(() => {
-      return remote.download();
-    })
-    .then((tempFolder) => {
-      return fs.copy(tempFolder, CACHE_FOLDER);
-    })
-    .then(() => {
-      return index.rebuildPagesIndex();
-    })
-    .then(() => {
-      return done();
-    })
-    .catch((err) => {
-      return done(err);
-    });
+  return fs.ensureDir(CACHE_FOLDER)
+    .then(() => remote.download())
+    .then((tempFolder) => fs.copy(tempFolder, CACHE_FOLDER))
+    .then(() => index.rebuildPagesIndex());
 };
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -23,7 +23,9 @@ exports.getPage = (page) => {
       let filePath = path.join(CACHE_FOLDER, 'pages', folder, page + '.md');
       return fs.readFile(filePath, 'utf8');
     })
-    .catch((err) => console.error(err));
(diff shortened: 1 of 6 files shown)


vladimyr:
Took one more step 😉 This branch now fully addresses #160 - promisified all remaining modules (`cache` & `tldr`) as part of 950e202.

lib/cache.js
```diff
@@ -9,47 +9,27 @@ const index = require('./index');
 
 const CACHE_FOLDER = path.join(config.get().cache, 'cache');
 
-exports.lastUpdated = (done) => {
-  fs.stat(CACHE_FOLDER, done);
-};
+exports.lastUpdated = () => fs.stat(CACHE_FOLDER);
```
agnivade: Please use the complete `() => {}` syntax everywhere for consistency.

lib/index.js
```diff
@@ -97,69 +77,32 @@ function clearRuntimeIndex() {
...
-            .catch((err) => {
-              reject(err);
-            });
-        } else {
-          return resolve(idx);
-        }
-      });
-  });
+  return fs.readJson(shortIndexFile)
+    .catch(() => {
```
agnivade: Please move the catch block after `then()`.

lib/tldr.js
```diff
@@ -15,16 +15,12 @@ const exit = process.exit;
 exports.list = (singleColumn) => {
   let os = platform.getPreferredPlatformFolder();
   index.commandsFor(os)
-    .then((commands) => {
-      printPages(commands, singleColumn);
-    });
+    .then((commands) => printPages(commands, singleColumn));
```
agnivade: Please use the full `() => {}` syntax as mentioned earlier.

lib/tldr.js
```diff
@@ -138,46 +108,40 @@ function printPages(pages, singleColumn) {
...
-  cache.getPage(command, (err, content) => {
-  /* eslint-enable */
-    if (!content) {
+  cache.getPage(command)
+    .then((content) => {
+      // If found in first try, render it
+      if (content) {
+        checkStale();
+        renderContent(content, options);
+        return exit(0);
```
agnivade: Do we need the `exit(0)` ? Won't it automatically exit with 0 ?

lib/tldr.js
```diff
@@ -193,13 +157,14 @@ function renderContent(content, options={}) {
 }
 
 function checkStale() {
-  cache.lastUpdated((err, stats) => {
-    if (err) {
+  cache.lastUpdated()
+    .then(stats => {
```
agnivade: Please use `()` around `stats`. Applies everywhere.

test/cache.spec.js
```diff
@@ -9,18 +9,16 @@ const platform = require('../lib/platform');
 
 
 describe('Cache', () => {
-  it('should return a positive number on lastUpdate', function(done) {
-    /* eslint-disable */ // To allow setting timeout of 30 secs
+  it('should return a positive number on lastUpdate', function () {
+    // To allow setting timeout of 30 secs
+    // eslint-disable-next-line no-magic-numbers
```
agnivade: This `eslint-disable-next-line` is very cool. Thanks.

lib/tldr.js
vladimyr: Ok, tried to go along with established conventions but missed this one 😞 
IMO it should be checked using eslint: https://eslint.org/docs/rules/arrow-parens

lib/index.js
vladimyr: Actually can't do that here because this `catch()` resolves with created index that gets consumed by next `then()`; think of it as async else block.

lib/tldr.js
vladimyr: You are referring to optional argument?
https://nodejs.org/dist/latest-v8.x/docs/api/process.html#process_process_exit_code

lib/tldr.js
vladimyr: Why here? Are you trying to completely avoid short form? 🤔

agnivade: Yes, that's right.

lib/tldr.js
agnivade: I meant, why do we need to call the exit() function at all ? Won't it exit normally by itself ?

lib/tldr.js
agnivade: Fully agree. Possibly something as a next PR ? :smile:

lib/tldr.js
vladimyr: Huh, luckily my whole PR is (not) going in opposite direction 😿 
I'll need more time to address all occurrences... 

Also this is good candidate for eslint config update: https://eslint.org/docs/rules/arrow-body-style

lib/tldr.js
vladimyr: Sure 😉

lib/tldr.js
vladimyr: In this case I'm actually terminating promise chain prematurely so it is required.

agnivade: Got it.

lib/tldr.js
vladimyr: Since you just merged #200 I'll merge master here to get ESLint guide me 💯 🏄

agnivade: Yep, that was the intention :laughing:

PR #203, opened 2018-03-23
Add cli spinners

vladimyr:
## Description
Use [`ora`](https://npm.im/ora) powered spinners instead of console logging. :tada:

Check it out in action:

[asciicast](https://asciinema.org/a/l42YYvpas5Jl5IX7MpcsD0PWo)

## Checklist

- [x] Code compiles correctly


lib/tldr.js
```diff
@@ -140,15 +145,16 @@ function printBestPage(command, options={}) {
         return exit();
       }
       // If not found, try to update
-      console.log('Page not found. Updating cache ..');
+      spinner.start('Page not found, updating cache');
```
agnivade: I would prefer to keep 2 sentences. I think grammatically that is more correct. You can remove the ellipses (`...`)

vladimyr: `ora` style is without any suffix so I tried to go with that which made me change from period to colon. :grin:
So you ar basicly saying lets use:
`Page not found. Updating cache`?

agnivade: Correct.

agnivade:
Thanks !

PR #205, opened 2018-03-31
Bugfix: Don't set shortIndex if index is empty

vladimyr:
## Description
Commit 84eb2bf introduced my _reviewed_ promisification code which unfortunately behaves incorrectly. While doing refactor I've accidentally changed behavior of `#readShortPagesIndex` to **always** assign `shortIndex` variable to calculated value which is in fact wrong in cases where `#buildShortPagesIndex` returns empty index. This leads to situation where very first query on initial run after installation (cache has not been downloaded yet) will fail because first check will assign `shortIndex` to empty object and after cache is downloaded client will ignore it and simply return `shortIndex`'s empty value due to:
```js
function getShortIndex() {
  if (shortIndex) {
    return Promise.resolve(shortIndex);
  }
  return readShortPagesIndex();
}
```
It is time to clean my mess, so that's why I'm offering this hotfix :ambulance: 

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -100,6 +100,10 @@ function getShortIndex() {
 // and load it into memory.
 function readShortPagesIndex() {
   return fs.readJson(shortIndexFile)
+    .then((idx) => {
+      shortIndex = idx;
+      return shortIndex;
+    })
     .catch(() => {
       // File is not present; we need to create the index.
       let idx = buildShortPagesIndex();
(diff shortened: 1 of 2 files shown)


agnivade:
Thanks. Can you add a test for this ?

vladimyr:
> Thanks. Can you add a test for this ?

I'll give it a shot... :pray:

vladimyr:
I'm afraid I don't know how to do it :man_facepalming:

vladimyr:
This is best I can do having in mind that `shortIndex` is inaccessible from outside: 1a29efb
Feels bit hacky :crying_cat_face:

agnivade:
Hmm .. that's right. I guess this has to do for now. I will see if it can be improved later.

PR #210, opened 2018-04-21
Use random temp folder for remote zip extraction

### LINKED ISSUE #208: Improve multi-user usage
## Expected behaviour
Should allow more than one user to use tldr

## Actual behaviour
When attempting to update the cache, it fails due to using a hardcoded path if another user has run tldr before you. This also occurs on first run because it hasn't setup local cache or config.

## Log, debug output
```
Updating{ Error: EACCES: permission denied, rmdir '/tmp/tldr/pages'
  errno: -13,
  code: 'EACCES',
  syscall: 'rmdir',
  path: '/tmp/tldr/pages' }
```
## Environment
- Operating system - Linux
- NodeJS version - 6.9.1

vladimyr:
## Description

This resolves #208 by using on-fly generated temporary folder path plus doing necessary cleanup after cache has been updated.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -13,7 +13,6 @@ const CACHE_FOLDER = path.join(config.get().cache, 'cache');
 const TEMP_FOLDER = path.join(os.tmpdir(), 'tldr');
 
 exports.CACHE_FOLDER = CACHE_FOLDER;
-exports.TEMP_FOLDER = TEMP_FOLDER;
 
 exports.lastUpdated = () => {
   return fs.stat(CACHE_FOLDER);
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -43,7 +43,7 @@ exports.update = () => {
 
   // Downloading fresh copy
   return Promise.all([
-    // Create new temporay folder
+    // Create new temporary folder
     fs.ensureDir(tempFolder),
     fs.ensureDir(CACHE_FOLDER)
   ])
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -12,6 +12,9 @@ const utils = require('./utils');
 const CACHE_FOLDER = path.join(config.get().cache, 'cache');
 const TEMP_FOLDER = path.join(os.tmpdir(), 'tldr');
 
+exports.CACHE_FOLDER = CACHE_FOLDER;
+exports.TEMP_FOLDER = TEMP_FOLDER;
+
 exports.lastUpdated = () => {
   return fs.stat(CACHE_FOLDER);
 };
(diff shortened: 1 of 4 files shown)


lib/remote.js
```diff
@@ -3,44 +3,61 @@
...
-// Downloads the zip file from github and extracts it to /tmp/tldr
+let request = require('request');
+
+function createTempFolder(prefix = 'tldr') {
+  let tmpdir = path.join(os.tmpdir(), prefix);
+  let target = path.join(tmpdir, utils.uniqueId());
+
+  return fs.mkdirp(target).then(() => {
+    return {
+      get path() {
```
agnivade: It is not immediately apparent why we are returning an object like this. Add a comment ?

lib/remote.js
```diff
@@ -3,44 +3,61 @@
...
   let url = config.get().repository;
-  let target = path.join(os.tmpdir(), 'tldr');
 
-  // Empty the tmp dir
-  return fs.emptyDir(target)
-    .then(() => {
-      // Creating the extractor
-      let extractor = unzip.Extract({ path: target });
+  // Create temp folder
+  return createTempFolder().then((tempFolder) => {
```
agnivade: Perhaps you can pass the `target` as an argument to `createTempFolder`. That would make it more testable.

vladimyr:
@agnivade I haven't addressed your comments directly because I've changed approach, however you asked for tests and here they are :wink:

lib/cache.js
```diff
@@ -33,15 +39,31 @@ exports.clear = () => {
...
-    .then((tempFolder) => {
+    .then(() => {
+      // Copy data to cache folder
       return fs.copy(tempFolder, CACHE_FOLDER);
     })
     .then(() => {
-      return index.rebuildPagesIndex();
+      return Promise.all([
+        // Remove temporary storage
+        fs.remove(TEMP_FOLDER),
```
agnivade: It seems that you are creating `tempFolder` inside `TEMP_FOLDER`. But you are deleting `TEMP_FOLDER`. This will fail if 2 users are simultaneously updating their caches. Since the tool will delete the parent dir, there might be a race condition where one user's cache dir gets deleted.

I would suggest to delete `tempFolder` instead and keep the parent dir as is.

vladimyr: I'm idiot, didn't take concurrency into equation 🤦♂
Good catch, will do as you suggested 👌

agnivade: That's what reviews are for :wink:

vladimyr:
> I would suggest to delete tempFolder instead and keep the parent dir as is.

Now, it should work as expected.

agnivade:
Thanks a bunch !

PR #212, opened 2018-04-27
[Update]: package-lock.json dependencies

### LINKED ISSUE #211: Vulnerability: hoek should be updated to 5.0.3
The package `hoek` is specified in the `package-lock.json` file at version `4.2.0`. When I run `npm-remote-ls hoek` though, it tells me that version 5.0.3 will suffice. I suggest updating it to mitigate the vulnerability that GitHub has identified.

Here's the reverse-dependency tree generated by `npm-remote-ls`: https://hastebin.com/enirizinan

agnivade:
Fixes #211

## Description
Update package-lock.json dependency

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,10 +4,19 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@sinonjs/formatio": {
+      "version": "2.0.0",
+      "resolved": "http://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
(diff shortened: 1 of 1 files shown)


agnivade:
@vladimyr - Can you take a look at the CI failure ? Node stable is failing due to some strange cause. The Node 7 failure looks legit.

vladimyr:
Node 7 thing is indeed legit and I just opened PR to address it. It was due to race condition in test itself which is semi-good thing because only test was problematic and actual behavior of client wasn't compromised. I hope that after this fix we won't see it again :crossed_fingers: 

Regarding node stable one, I just eyeballed logs on travis and it looks weird :dizzy_face: Do you want me to investigate it further?

PR #215, opened 2018-05-01
[travis]: re-introduce the stable version

agnivade:


agnivade:
package-lock.json isn't fetched while installing a package. So won't work.

PR #216, opened 2018-05-03
Making tldr compatible with node 10

vladimyr:
## Description
Swapping `unzip2` with [`node-unzip-2`](https://github.com/glebdmitriew/node-unzip-2) fork (uses `graceful-fs@4`) in order to avoid node 10 core dump.
Source: https://github.com/nodejs/node/issues/20285#issuecomment-385299321

## Checklist

- [x] Code compiles correctly


vladimyr:
@agnivade There were 2 different things that made node 10 fail:
1. old version of `unzip2` package that used problematic `graceful-fs@3` under the hood causing core dump
2. caching `node_modules` folder on subsequent installs on travis

Second is problematic because `webworker-threads` needs to be rebuilt when you go from node 9 to node 10. So we need to either stop caching `node_modules` between different environments/installs or make sure that we call `npm rebuild` before we run tests. First option makes more sense to me but if you want it other way around let me know :wink:

agnivade:
> caching node_modules folder on subsequent installs on travis

Ah so that was the issue ! I was investigating but could not figure out the webworker failures. I have a different approach without changing the unzip2 dependency here at #215, which I believe is slightly more safe than changing the dependency altogether.

Let me know what you think.

vladimyr:
Hm, I also bumped `natural` version but it doesn't really change anything because it did not cause failures in the first place... Am I missing something? :thinking: 

Regarding dependency change, it is actually exactly same codebase just with updated deps:
1. [`unzip2@0.2.5`](https://unpkg.com/unzip2@0.2.5/lib/) published 4 yrs ago using `graceful-fs@3` through `fstream@0.1.x`
2. [`node-unzip-2@0.2.7`](https://unpkg.com/node-unzip-2@0.2.7/lib/) published few months ago using `graceful-fs@4` through `fstream@1.0.x`

so technically it is a depedency change but what it does it just updates dependencies of `unzip2` so we can get rid of problematic `graceful-fs@3` that caused node 10 to core dump.

vladimyr:
Now I'm confused :dizzy: 
Checked your branch, and it really does not have `graceful-fs@3` inside `package-lock.json` which is superweird due to following:
https://github.com/tldr-pages/tldr-node-client/blob/240f95740f83a5c576bb1814084ef0425af395ea/package-lock.json#L1897-L1903
https://github.com/tldr-pages/tldr-node-client/blob/240f95740f83a5c576bb1814084ef0425af395ea/package-lock.json#L743-L748
But in the same time, [`package.json`](https://github.com/npm/fstream/blob/3512b3d41b4c9b3ef15cf32a04e06b297f3dd6a4/package.json) of `fstream@0.1.31` says that it requires `"graceful-fs": "~3.0.2"`...

agnivade:
> Did you change package-lock.json manually 

Yes, I did :laughing: 

> package-lock.json which isn't normally published on npm.

Oh shucks .. didn't think of that. Ok, let us go with node-unzip2.

PR #223, opened 2018-05-13
Graceful fallback for webworker-threads import error

vladimyr:
## Description
Implementing safe import mechanism for native dependencies.

If execution of CLI fails due to [`webworker-threads` error](https://github.com/tldr-pages/tldr-node-client/blob/58cc72088dae7b206f1f0bb6d7404204584becb6/README.md#installation-issues), following message gets outputted:
error report message

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "6"
   - "7"
   - "8"
   - "9"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - "6"
   - "7"
   - "8"
   - "9"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,6 @@
 # Test against this version of Node.js
 environment:
   matrix:
-    - nodejs_version: "6"
     - nodejs_version: "7"
     - nodejs_version: "8"
     - nodejs_version: "9"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ environment:
     - nodejs_version: "6"
     - nodejs_version: "7"
     - nodejs_version: "8"
+    - nodejs_version: "9"
     - nodejs_version: Stable
 
 # Install scripts. (runs after repo cloning)
@@ -18,9 +19,6 @@ test_script:
   - node --version
   - npm --version
   - npm run test:all
-
-cache:
-  - node_modules
 
 # Don't actually build.
 build: off
--- a/lib/search.js
+++ b/lib/search.js
@@ -1,8 +1,22 @@
 'use strict';
 
+const chalk = require('chalk');
 const fs = require('fs-extra');
 const path = require('path');
-const natural = require('natural');
+const importSafe = require('./utils').importSafe;
+const natural = importSafe('natural', (err, query) => {
+  if (err.code !== 'ERR_REQUIRE_NATIVE') throw err;
+  const rootPath = path.join(__dirname, '../');
+  console.error(chalk`
+    {bgRed.whiteBright Error} Executing {bold.cyan tldr} failed due to incorrect native dependencies
+      target: {blue ${path.relative(rootPath, err.filename)}}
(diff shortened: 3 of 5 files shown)


vladimyr:
You can see effects of this change here:
https://ci.appveyor.com/project/igorshubovych/tldr-node-client-bnut4/build/job/eue0lk1qtkye67og
because building on appveyor fails due to cached `node_modules`.

vladimyr:
I've reintroduced testing on `node@6` (on Travis) due to:
https://github.com/tldr-pages/tldr-node-client/blob/58cc72088dae7b206f1f0bb6d7404204584becb6/package.json#L20-L22
and fixed appveyor testing.

CI results:
- **travis** https://travis-ci.org/tldr-pages/tldr-node-client/builds/378405064
- **appveyor** https://ci.appveyor.com/project/igorshubovych/tldr-node-client-bnut4/build/1.0.433

vladimyr:
Hm, I could but what do we get from dropping `node@6` support? I mean why not support it if it doesn't cost us anything? 🤔

agnivade:
We would have to keep adding new node versions to the travis and appveyor files, every time a new one gets released. This would keep bloating up the files unnecessarily. So either we remove 6 from CI tests, or we upgrade to 7 in package.json.

On second thoughts, I think we should keep package.json as is, and make the CI versions 7,8,9,stable. i.e. - do not test against 6. If at any time, we see an issue specific to 6, then we can think of dropping support for 6, and bumping up package.json. Until then, let's not test against 6.

vladimyr:
If I understand you correctly, you are suggesting that support for `node@6` should be _silently_ dropped (without communicating that to users by altering `package.json`) and that could become official if someone comes in with `node@6` specific bug? You are basically relaying on historical fact that app is already well tested there so probability of _new_ `node@6` is really low?

I understand your incentive but I'm not 100% percent thrilled with non-transparent support drop. However, I agree that chances of bad outcome are low so I'm kicking `node@6` out of test matrix right away...

agnivade:
> If I understand you correctly, you are suggesting that support for node@6 should be silently dropped (without communicating that to users by altering package.json)

Yes I was suggesting that. But then in the second para, changed my mind. :)

Keeping the test matrix to just the latest 4 versions looks good to me.

agnivade:
Going through the PR now - I would rather let npm deal with the installation issue, than us trying to implement fallback mechanism when something has failed. Writing import checks like this from within the code feels hackish to me. And also, in systems where the installation works fine, the code becomes redundant.

vladimyr:
Right now code contains:
- `package.json` that advertises `node@6` support
- CI services setup that enables testing on `7, 8, 9` and `stable (10)`

vladimyr:
> Going through the PR now - I would rather let npm deal with the installation issue, than us trying to implement fallback mechanism when something has failed.

It is actually just pretty report done through require hook and original error (behavior) isn't altered.

> Writing import checks like this from within the code feels hackish to me.

As much as monkey-patching functions is :grin: 

> And also, in systems where the installation works fine, the code becomes redundant.

This is very true but it is also a good thing. Only cost that gets payed is visual clutter, no performance penalty and in cases where it goes wrong users are provided with nicer (less scary) error message that points them in right direction...

vladimyr:
> I don't understand. Are you saying that currently, some functions are being monkey-patched ?

I was talking about code in this PR. In order to hook into require system I had to patch `require.extensions['.node']`...

> It's true that there is no performance penalty. But some extra lines of code to handle something which I feel is outside of a module's purview, seems like incurring a cost to me. Overall, I don't see a good reason to add this in the codebase.

I proposed it in order to improve user experience when issue arises and that might happen more often now because node 9 to 10 switch requires recompilation. That being said it is messy because playing with node internals requires it to be so. So it is a question of _user experience_ vs _developer experience_. If you think that handling this edge case isn't really justifiable and/or cons are bigger than pros feel free to close this PR or let me know to close it myself, no hard feelings :smiley:

agnivade:
Thanks. I would still like to get in the appveyor.yml changes. It would be great if you can update the PR to just reflect the changes in appveyor.yml.

vladimyr:
I'll open separate PR for that and close this one.

PR #243, opened 2018-12-24
Make user messages more explicit

waldyrious:
Just small wording and punctuation adjustments to make user-facing messages more explicit.


agnivade:
Thanks !

PR #249, opened 2019-02-08
Fix npm vulnerabilities

sbrl:
This fixes the npm vulnerabilities that I was notified about, by running `npm audit fix`.

The only slight issue remaining is that the `text-encoding` package isn't maintained anymore:

```
npm WARN deprecated text-encoding@0.6.4: no longer maintained
```

However, according to `npm-why`, this may not be our problem:

```
$ npm-why text-encoding
Who required text-encoding:

  tldr > sinon > nise > text-encoding@0.6.4
```

Then again, if the `sinon` package version we're requiring isn't the latest version, it could be locked to an older version of `nise`, etc.

The `npm-check` command (also installable via `npm`) can come in handy here. It checks whether the dependencies you have installed are up-to-date, and provides an easy text-based UI to update said packages. There are a bunch of other updates said tool is suggesting, but here's the one we're probably interested in:

```
......
sinon        😎  MAJOR UP  Major update available. https://sinonjs.org/
                          npm install --save-dev sinon@7.2.3 to go from 4.5.0 to 7.2.3

Use npm-check -u for interactive update.
```

This PR does not update this `sinon` package - that would be a separate PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,15 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@sinonjs/commons": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.3.0.tgz",
+      "integrity": "sha512-j4ZwhaHmwsCb4DlDOIWnI5YyKDNMoNThsmwEpfHx6a1EpsGZ9qYLxP++LMlmBRjtGptGHFsGItJ768snllFWpA==",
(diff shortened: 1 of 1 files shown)


vladimyr:
Hmm, this is fine but will remove warnings only if you install it in _development mode_. `package-lock` isn't published (and can't be) so end users will/might still get warnings on install.

sbrl:
@vladimyr I see. I guess at some point we'll need to investigate fully updating all the dependencies to the very latest versions.

PR #264, opened 2019-06-18
Housekeeping

vladimyr:
## Description

Basically this boils down to multiple small behavior non-altering changes:
- renaming `bin/tldr` to `bin/cli.js` because it is js file after all :)
- swapping ESLint's magical numbers with string counterparts
- updating travis config & minimal supported version inside `package.json`
- fixing editorconfig issues (spacing, newlines...)
- tweaking `package.json` scripts
- updating `main` to point to actual library used by cli

## Checklist

- [x] Code compiles correctly


test/render.spec.js
```diff
@@ -102,7 +102,7 @@ describe('Render', () => {
       'theme': 'bad'
     });
 
-    let spy = sinon.spy(console, 'error');
+    let spy = sinon.stub(console, 'error');
```
vladimyr: Without this `test:quiet` isn't really quiet :upside_down_face:

agnivade:
Thanks @vladimyr. Could you please split unrelated changes into different PRs ? Otherwise the commits and then the review cycles get messed up.

We can get the .eslintrc changes in one, and the README and config.js changes in another.

Regarding the other changes:
- I wouldn't want to rename bin/tldr to bin/cli.js. Mainly because when we run a command, we run it as `tldr <page>`, not `tldr.js <page>`. So it maintains consistency with the final command.
- Let us not add version 8 into .travis.yml and change the engines version to 8. The package.json field is just a warning, and I would want to keep it as is unless we introduce a breaking change. But that does not necessarily mean we _have_ to test every version from 6 onwards. Let us just keep testing stable and the last 3 releases.

agnivade:
@vladimyr - Would you be able to take a look at addressing some comments that I had ?

vladimyr:
> We can get the .eslintrc changes in one, and the README and config.js changes in another.

Ok, that makes sense. I'll close this PR and split it into multiple smaller ones.

> * I wouldn't want to rename bin/tldr to bin/cli.js. Mainly because when we run a command, we run it as `tldr <page>`, not `tldr.js <page>`. So it maintains consistency with the final command.

My response to this one is c/p straight from my terminal:
```
➜ realpath `which npm`
/Users/vladimyr/.fnm/node-versions/v10.17.0/installation/lib/node_modules/npm/bin/npm-cli.js
```
The vast majority of node CLIs out there follow the same pattern and I honestly don't know what makes you think it is good to go in opposite direction backed by given explanation. Common conventions might not be perfect but they are common.

> * Let us not add version 8 into .travis.yml and change the engines version to 8. The package.json field is just a warning, and I would want to keep it as is unless we introduce a breaking change. But that does not necessarily mean we _have_ to test every version from 6 onwards. Let us just keep testing stable and the last 3 releases.

I very much disagree with this one. You can't really stitch your own perceived meaning to something. `engines` field isn't fingers crossed informal love letter to the user but: https://docs.npmjs.com/files/package.json#engines
>You can specify the version of node that your stuff **works** on:

If you advertise something without testing then it is not a matter if it will go sideways but when!
I see two possibilities here:
1. Either we keep stuff as is but test those older versions (LTS only!)
2. We go and follow others by stating v8+ only

agnivade:
Sorry I am not very well-versed in the npm conventions, but I agree with your thoughts.

> The vast majority of node CLIs out there follow the same pattern and I honestly don't know what makes you think it is good to go in opposite direction backed by given explanation. 

Yes I see. Let's go with that then. We would also then need to setup the symlink linking `tldr` to `tldr.js` ?

> If you advertise something without testing then it is not a matter if it will go sideways but when!

My only concern was that if we do 1, then the list of versions we test against will keep on increasing with every new release. If we go with 2, then we have to keep updating `engines` after every new release. Do you have any idea on what other npm packages do normally ?

vladimyr:
> Yes I see. Let's go with that then. We would also then need to setup the symlink linking `tldr` to `tldr.js` ?
> 
> >

`npm` (and all other js pm-s) do that for you automatically according to `bin` field inside `package.json`:
https://npmfs.com/package/npm/6.14.4/package.json#L32

> My only concern was that if we do 1, then the list of versions we test against will keep on increasing with every new release. If we go with 2, then we have to keep updating `engines` after every new release.

Which is very true and here is what others do: https://github.com/eslint/eslint/blob/master/CHANGELOG.md Search for "drop Node.js". So, yeah, I think we should do 2nd approach and conservatively drop version support so it won't incur frequent `engines` changes.

agnivade:
Sounds good to me. Thanks @vladimyr !

PR #269, opened 2019-07-04
Refactor: change cache and tldr to be a class

agnivade:
## Description

- This fixes global config settings from bin/tldr and localizes it inside tldr.
- It directly allows creating the cache object as a class member.
- We also remove the unused platform.resolve function.
- And since config is a class member of tldr, it also allows us to pass it directly to render.toANSI function
removing yet another global config invocation.
- Use objects of cache class everywhere. This reuses the cacheFolder path
and also keeps state localized. And also reduces API surface.

## Checklist

- [x] Code compiles correctly
~- [ ] Created tests, if possible~

### DIFF
--- a/lib/tldr.js
+++ b/lib/tldr.js
@@ -4,7 +4,7 @@ const sample = require('lodash/sample');
 const fs = require('fs-extra');
 const ms = require('ms');
 const ora = require('ora');
-const cache = require('./cache');
+const Cache = require('./cache');
 const search = require('./search');
 const platform = require('./platform');
 const messages = require('./messages');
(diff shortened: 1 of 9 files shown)


agnivade:
@vladimyr, want to take a look ?

lib/tldr.js
```diff
@@ -13,188 +13,200 @@ const render = require('./render');
...
+    let os = platform.getPreferredPlatformFolder(this.config);
+    index.commandsFor(os)
+      .then((pages) => {
+        if (pages.length === 0) {
+          console.error(messages.emptyCache());
+          exit(1);
+        }
+        let page = sample(pages);
+        console.log('PAGE', page);
+        this.printBestPage(page, options);
```
sbrl: Because this is a reference to `this` inside a `.then()`, does the arrow function need to be bound like this?

```js
// ....
.then((() => {
    // ....
}).bind(this);
```

Sometimes it does and sometimes it doesn't - I'm actually rather unclear on how it works here.

In some situations I find that `this` refers to the global scope instead of the class instance, leading to errors.

agnivade: It doesn't. `this` inside an arrow function does not redefine it. It just uses the parent `this`. https://exploringjs.com/es6/ch_arrow-functions.html#sec_traditional-functions-bad-non-methods

sbrl: Oooh, I didn't know that arrow functions have different scoping rules. I'll have to play around with  that. Fascinating!

agnivade:
Merging these. If anyone has any comments, feel free to post them. I can send future PRs.

PR #270, opened 2019-07-07
Added: language support

agnivade:
## Description
An environment variable LANG needs to contain the correct BCP-47 tag
for it to pick up the right language file. :tada: 

## Checklist

- [x] Code compiles correctly

@sbrl @mebeim @pxgamer @waldyrious

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To see tldr pages:
 - `tldr <command> --os=<platform>` show command page for the given platform (`linux`, `osx`, `sunos`)
 - `tldr --search "<query>"` search all pages for the query
 - `tldr --linux <command>` show command page for Linux
-- `tldr  --osx <command>` show command page for OSX
+- `tldr --osx <command>` show command page for OSX
 - `tldr --sunos <command>` show command page for SunOS
 - `tldr --list` show all pages for current platform
 - `tldr --list-all` show all available pages
@@ -58,6 +58,10 @@ There are more commands to control the local cache:
 As a contributor, you might also need the following commands:
 
 - `tldr --render <path>` render a local page for testing purposes
+
+To view tldr pages for a different language, set an environment variable `LANG` containing a valid [BCP-47](https://tools.ietf.org/html/bcp47) code (such as `zh`, `pt-BR`, or `fr`) and then run the above commands as usual. For the list of available translations, please refer to the main [tldr](https://github.com/tldr-pages/tldr) repo.
+
+- `LANG=zh tldr <command>`
 
 ## Configuration
 
@@ -147,18 +151,18 @@ fpath = (my/completions $fpath)
 
 #### Installation Issues
 
-- If you are trying to install as non-root user (`npm install -g tldr`) and get something like - 
- 
+- If you are trying to install as non-root user (`npm install -g tldr`) and get something like -
+
  ```
  Error: EACCES: permission denied, access '/usr/local/lib/node_modules/tldr'
  ```
- 
+
  Then most probably your npm's default installation directory has improper permissions. You can resolve it by clicking [here](https://docs.npmjs.com/getting-started/fixing-npm-permissions)
- 
-- If you are trying to install as a root user (`sudo npm install -g tldr`) and get something like - 
+
+- If you are trying to install as a root user (`sudo npm install -g tldr`) and get something like -
 
 ```
-as root -> 
+as root ->
 gyp WARN EACCES attempting to reinstall using temporary dev dir "/usr/local/lib/node_modules/tldr/node_modules/webworker-threads/.node-gyp"
 gyp WARN EACCES user "root" does not have permission to access the dev dir "/usr/local/lib/node_modules/tldr/node_modules/webworker-threads/.node-gyp/8.9.1"
 ```
--- a/README.md
+++ b/README.md
@@ -59,7 +59,9 @@ As a contributor, you might also need the following commands:
 
 - `tldr --render <path>` render a local page for testing purposes
 
-To view tldr pages for a different language, set an environment variable `LANG` containing a valid [BCP-47](https://tools.ietf.org/html/bcp47) code (such as `zh`, `pt-BR`, or `fr`) and then run the above commands as usual. For the list of available translations, please refer to the main [tldr](https://github.com/tldr-pages/tldr) repo.
+To view tldr pages for a different language, set an environment variable `LANG` containing a valid [POSIX locale](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) (such as `zh`, `pt_BR`, or `fr`) and then run the above commands as usual. In most `*nix` systems, this variable will already be set.
+
+For the list of available translations, please refer to the main [tldr](https://github.com/tldr-pages/tldr) repo.
 
 - `LANG=zh tldr <command>`
 
--- a/README.md
+++ b/README.md
@@ -59,11 +59,11 @@ As a contributor, you might also need the following commands:
 
 - `tldr --render <path>` render a local page for testing purposes
 
-To view tldr pages for a different language, set an environment variable `LANG` containing a valid [POSIX locale](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) (such as `zh`, `pt_BR`, or `fr`) and then run the above commands as usual. In most `*nix` systems, this variable will already be set.
-
-For the list of available translations, please refer to the main [tldr](https://github.com/tldr-pages/tldr) repo.
+Tldr pages defaults to showing pages in the current language of the operating system, or English if that's not available. To view tldr pages for a different language, set an environment variable `LANG` containing a valid [POSIX locale](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) (such as `zh`, `pt_BR`, or `fr`) and then run the above commands as usual. In most `*nix` systems, this variable will already be set.
 
 - `LANG=zh tldr <command>`
+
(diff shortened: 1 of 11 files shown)


agnivade:
Sure

lib/cache.js
```diff
@@ -22,12 +22,16 @@ class Cache {
 
   getPage(page) {
     let preferredPlatform = platform.getPreferredPlatformFolder(this.config);
-    return index.findPlatform(page, preferredPlatform)
+    let preferredLanguage = 'en';
+    if (process.env['LANG'] !== '') {
+      preferredLanguage = process.env['LANG'];
```
mebeim: Hrm, this is probably not what you would want. Our client specification does not follow [the POSIX locale specification](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) for language environment variables. As a result of this, the `LANG` environment var is basically never going to be right, and should be converted first.

Some examples in case to make it clear:

- `LANG="it_IT.UTF-8"` and `LANG="it_IT"` should be converted to `it-IT`, then check for `pages.it-IT/platform/page.md` or `pages.it/platform/page.md`.
- `LANG="en_US.UTF-8"` and `LANG="en_US"` should be converted to `en-US`, then (special case since language code is `en`) check for `pages/platform/page.md`.
- `LANG="en.UTF-8"` and and `LANG="en"` should be converted to `en`, then (special case since language code is `en`) check for `pages/platform/page.md`.
- `LANG="sr_RS@latin"` should be converted to `sr-RS`, then check for `pages.sr-RS/platform/page.md` or `pages.sr/platform/page.md`.

---

This is an issue I've been thinking about for quite a while. I really think we should update the client spec and follow the POSIX locale specification (basically just use `ll_CC` instead of `ll-CC`). It would make it really easier to deal with languages.

Anyway, the fallback to `ll` from `ll_CC` (or `ll-CC`) if `pages.ll_CC` (or `pages.ll-CC`) is not present should be made regardless.

pxgamer: I was thinking this earlier too. I know there was a reason that we chose the current language code format though, can't remember why.

mebeim: @pxgamer well, the only issue here is the dash `-` versus the underscore `_`. I don't remember a real reason for this choice, if I recall right :thinking:.

agnivade: It's here - https://gitter.im/tldr-pages/tldr?at=5c387be59289cc7aa757c1ad. You commented yourself @mebeim :) 

tldr is  GNU uses _ and BCP-47 uses -. They have different standards. And we decided to go with BCP-47.

But you are actually right anyways, I need to parse the BCP-47 tag first. Probably need to use a proper library.

README.md
```diff
@@ -57,6 +57,10 @@ As a contributor, you might also need the following commands:
 
 - `tldr --render <path>` render a local page for testing purposes
 
+To view tldr pages for a different language, set an environment variable `LANG` containing a valid [BCP-47](https://tools.ietf.org/html/bcp47) code. And then run the above commands as usual. Currently, we have translations for `zh`, `pt-BR`, `it`, `ta`, `fr`. For an updated list of translations, please refer to the main [tldr](https://github.com/tldr-pages/tldr) repo.
```
waldyrious: Instead of phrasing this as a list of existing translations, it would be easier to just give some examples of codes that can be used. That way there won't be a need to constantly update this list as translations in more languages are contributed. For example:

```suggestion
To view tldr pages for a different language, set an environment variable `LANG` containing a valid [BCP-47](https://tools.ietf.org/html/bcp47) code (such as `zh`, `pt-BR`, or `fr`) and then run the above commands as usual. For the list of available translations, please refer to the main [tldr](https://github.com/tldr-pages/tldr) repo.
```

agnivade: Sure

lib/cache.js
sbrl: Huh, this is a problem. I wasn't aware of what the POSIX standard was when I wrote the client spec. Might be worth tweaking that part of the spec?

mebeim: Thanks for the link @agnivade makes a lot more sense now, we were looking at BCP-47 and not POSIX. That's a bummer since now that I think about it the POSIX version is probably the only one you'll find used around when it comes to programming and environment variables :confused:.

agnivade: Sure. We would need to update documentation and rename the folders. Want to do that ?

mebeim: @agnivade I'm conflicted on whether or not to do it. Either way, this should be decided now before clients start supporting languages to create minimum trouble. I'll open an issue on the tldr repo so we can discuss this 🙂

sbrl: I'd advocate changing it now, since barely any clients have implemented language support. The longer we leave it, the more complicated it'll become.....

We could even require that clients support both :P

agnivade: Fixed.

agnivade:
@mebeim - This is ready for review.

@waldyrious - I updated the README text slightly. Please do check once to ensure everything looks good to you.

README.md
```diff
@@ -59,6 +59,12 @@ As a contributor, you might also need the following commands:
 
 - `tldr --render <path>` render a local page for testing purposes
 
+To view tldr pages for a different language, set an environment variable `LANG` containing a valid [POSIX locale](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) (such as `zh`, `pt_BR`, or `fr`) and then run the above commands as usual. In most `*nix` systems, this variable will already be set.
```
waldyrious: Maybe prefix this with information about what language is used by default. Something like this:

> tldr-pages defaults to showing pages in the current language of the operating system, or English if that's not available.

WDYT? (Assuming it accurately described the actual behavior, of course.)

README.md
```diff
@@ -59,6 +59,12 @@ As a contributor, you might also need the following commands:
...
+
+For the list of available translations, please refer to the main [tldr](https://github.com/tldr-pages/tldr) repo.
+
+- `LANG=zh tldr <command>`
```
waldyrious: I think it would be better (sequence-wise) to have this example before the previous sentence.

README.md
```diff
@@ -147,18 +153,18 @@ fpath = (my/completions $fpath)
 
 #### Installation Issues
 
-- If you are trying to install as non-root user (`npm install -g tldr`) and get something like - 
- 
+- If you are trying to install as non-root user (`npm install -g tldr`) and get something like -
```
waldyrious: I know it's not related to the changes in this PR, but the appropriate punctuation here would be `:`, not `-`. Since you're already editing these lines, maybe fix that too.

waldyrious:
> @waldyrious - I updated the README text slightly. Please do check once to ensure everything looks good to you.

Thanks for the ping. I left some minor comments inline, but overall it looks good. 👍

mebeim:
@agnivade does this PR follow the client spec? In particular, this:

> The [`LANG` environment variable](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html), if present, MUST be used to determine the language of pages to display.
>
> The [`LANGUAGE` environment variable](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html) specifies a priority list of languages that a user wishes to read in. In absence of `LANG` and if this environment variable present, a client MUST use the defined priority list to decide on the language of pages to display. If a page is not available in the user's preferred language, then a client MUST respect the user's priority list defined in the `LANGUAGE` variable, and MAY choose to notify the user that a page in their chosen language couldn't be found (perhaps along with a link to the [translations section of the contributing guide](https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md#translations)).

I'm sorry but I don't have enough time to review/test this in detail now.

agnivade:
It does not support the `LANGUAGE` or `LC_MESSAGES` yet. Only `LANG`. That is a TODO item for future.

README.md
agnivade: Sounds good.

README.md
```diff
@@ -59,6 +59,12 @@ As a contributor, you might also need the following commands:
 
 - `tldr --render <path>` render a local page for testing purposes
 
+Tldr pages defaults to showing pages in the current language of the operating system, or English if that's not available. To view tldr pages for a different language, set an environment variable `LANG` containing a valid [POSIX locale](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html#Locale-Names) (such as `zh`, `pt_BR`, or `fr`) and then run the above commands as usual. In most `*nix` systems, this variable will already be set.
+
+- `LANG=zh tldr <command>`
```
sbrl:
```suggestion
It is suggested that the `LANG` environment variable be set system-wide if this isn't already the case. Users without `sudo` access can set it locally in their `~/.profile`.

- `LANG=zh tldr <command>`
```

agnivade:
> I did think of an edge case where both aa_BB and aa exist, but aa has pages that aa_BB does not have, but I'm unsure as to how important it would be to handle that.

It is already handled. :)

```js
let ll;
      if (preferredLanguage.indexOf('_') > 0) {
        ll = preferredLanguage.substring(0, preferredLanguage.indexOf('_'));
      }
      if (!hasLang(targets, preferredLanguage)) {
        preferredLanguage = ll;
      }
```

lib/cache.js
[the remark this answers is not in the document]
agnivade: The format of `LANGUAGE` is slightly different. I plan to add support for these later.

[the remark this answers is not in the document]
sbrl: Ah, I didn't realise.

Yeah, let's do that in a future PR.

mebeim:
So @agnivade should this be squashed? Seeing the commits you are adding I would assume so.

agnivade:
Yep

agnivade:
Thanks guys, merging.

PR #284, opened 2019-11-26
Fix: Handle undefined LANG variable

### LINKED ISSUE #283: Error "Cannot read property 'includes' of undefined"
## Expected behaviour
It shows a man page

## Actual behaviour
It throws an error `Cannot read property 'includes' of undefined` and fails (see log below). Clean install, right after `npm i -g tldr`.

## Log, debug output
I use zsh:
```
tldr git-stash

TypeError: Cannot read property 'includes' of undefined
    at getShortIndex.then (/usr/local/lib/node_modules/tldr/lib/index.js:24:29)
✔ Page not found. Updating cache...
✔ Creating index...
TypeError: Cannot read property 'includes' of undefined
    at getShortIndex.then (/usr/local/lib/node_modules/tldr/lib/index.js:24:29)
Page not found.
```

## Environment
macOS Catalina 10.15.15
node v10.15.1

agnivade:
Fixes #283

## Checklist

~- [ ] Code compiles correctly~

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -22,10 +22,7 @@ class Cache {
 
   getPage(page) {
     let preferredPlatform = platform.getPreferredPlatformFolder(this.config);
-    let preferredLanguage = 'en';
-    if (process.env['LANG'] !== undefined && process.env['LANG'] !== '') {
-      preferredLanguage = process.env['LANG'];
-    }
+    const preferredLanguage = process.env.LANG || 'en';
     return index.findPage(page, preferredPlatform, preferredLanguage)
       .then((folder) => {
         if (!folder) {
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -23,7 +23,7 @@ class Cache {
   getPage(page) {
     let preferredPlatform = platform.getPreferredPlatformFolder(this.config);
     let preferredLanguage = 'en';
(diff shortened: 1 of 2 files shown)


lib/cache.js
```diff
@@ -23,7 +23,7 @@ class Cache {
   getPage(page) {
     let preferredPlatform = platform.getPreferredPlatformFolder(this.config);
     let preferredLanguage = 'en';
-    if (process.env['LANG'] !== '') {
+    if (process.env['LANG'] !== undefined && process.env['LANG'] !== '') {
```
vladimyr: 1. `process.env` holds dictionary of strings so it can be either a string, an empty string or `undefined`
2. because both empty string and `undefined` are falsey values you are supposed to check it in usual and most logical way:
```js
if (process.env['LANG']) {
  preferredLanguage = process.env['LANG'];
}
```
3. There isn't something magical about uppercase variable names in javascript so usual way of _dot access_ should be used: `process.env.LANG` 
4. This boils down to using `LANG` or predefined fallback which is basically very definition of OR assignment:
```js
let preferredLanguage = process.env.LANG || 'en';
```
5. And because it is never mutated it should be also marked as `const` which brings us to final version:
```js
const preferredLanguage = process.env.LANG || 'en';
```

I'll refrain from making any additional comments.

PR #287, opened 2019-12-30
Add missing keywords from the main repository

zdroid:
Source: tldr-pages/tldr description

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "man",
     "pages",
     "cheatsheets",
-    "examples,
+    "examples",
     "help",
     "unix",
     "linux",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -7,6 +7,10 @@
   "repository": "tldr-pages/tldr-node-client",
   "keywords": [
     "man",
+    "pages",
+    "cheatsheets",
+    "examples,
```
vladimyr: Closing quotes aren't optional 🙃

zdroid: Hvala!

vladimyr: I drugi put 😉

PR #291, opened 2020-04-14
Update cache if older than 30 days

briandk:
## Overview

This PR supersedes #198.

The previous behavior would still check if the cache were more than 30 days out of date, but all it would do was warn the user to run `tldr --update`.

The intention behind this fix is that because tldr CAN update itself, it should just try to update automatically when the cache is out of date instead of bothering the user.

## Description

- There is now a `cache.isStale()` method which returns a boolean indicating whether or not the cache is more than 30 days out of date
- `checkStale()` in `tldr.js` now calls `cache.isStale()`, and if that call returns true, it automatically runs `updateCache` instead of warning the user.

## Checklist

- [X] Code compiles correctly
- [X] Created tests, if possible
- [X] All tests passing (`npm run test:all`)
- [X] Extended the README / documentation, if necessary

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const fs = require('fs-extra');
+const ms = require('ms');
 const os = require('os');
 const path = require('path');
 const remote = require('./remote');
@@ -18,6 +19,18 @@ class Cache {
 
   lastUpdated() {
     return fs.stat(this.cacheFolder);
+  }
+
+  isStale() {
+    this
+      .lastUpdated()
+      .then((stats) => {
+        return stats.mtime < Date.now() - ms('30d');
+      })
+      .catch((err) => {
+        console.error(err);
+        process.exit(1);
+      });
   }
 
   getPage(page) {
--- a/lib/tldr.js
+++ b/lib/tldr.js
@@ -2,7 +2,6 @@
 
 const sample = require('lodash/sample');
 const fs = require('fs-extra');
-const ms = require('ms');
 const ora = require('ora');
 const Cache = require('./cache');
 const search = require('./search');
@@ -97,7 +96,8 @@ class Tldr {
   updateCache() {
     const spinner = ora();
     spinner.start('Updating...');
-    return this.cache.update()
+    return this.cache
+      .update()
       .then(() => {
         spinner.succeed();
       })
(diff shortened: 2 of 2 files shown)


agnivade:
Awesome, thanks ! I will get to this soon.

vladimyr:
@briandk To add on @agnivade's previous comment I propose you do `lastUpdated` calculation eagerly by using sync stat call. Then the whole `tldr#checkStale` method becomes sync.

Still, this is only a half solution because what exactly happens if the cache is stale but contains the page I asked for? According to code, it will show me a possibly outdated page and update the cache in the background. I would argue that it is not the nicest used experience. What needs to be done is:
1. check if the cache is stale
2. if it is, ask the user if they like to update before rendering the page
3. pull the page from cache and render it
4. do not try to update cache afterward if they didn't ask us to do it

Also because that makes page rendering potentially interactive process there should be a corresponding config flag that disables asking for an update and swaps it with current behavior (info message + manual update afterward) to preserve non-interactive (scriptable) behavior.

A great example of how things _should_ work is oh-my-zsh's auto-update feature.

agnivade:
> According to code, it will show me a possibly outdated page and update the cache in the background.

After we make the call sync, won't it show the page only after updating the cache.

vladimyr:
> After we make the call sync, won't it show the page only after updating the cache.

Only the `checkStale` in its current state, not check & update combo.

agnivade:
> Only the checkStale in its current state, not check & update combo.

Yes of course, we need to make the entire chain linked. Sorry for not being clear.

agnivade:
@briandk - Any thoughts on this ?

briandk:
So, @vladimyr @agnivade: your proposal is:

> 
> 1. check if the cache is stale
> 2. if it is, ask the user if they like to update before rendering the page
> 3. pull the page from cache and render it
> 4. do not try to update cache afterward if they didn't ask us to do it

My question is: why do we ask the user whether they want to update the cache? Are there ever actually situations where the answer is "no, I don't want to update the cache"?

What would be the problem if we did

1. Check if the cache is stale
2. If it is, **update the cache**
3. Pull the page from cache and render it

Honest question, by the way. I'm trying to learn 😇

vladimyr:
> My question is: why do we ask the user whether they want to update the cache? Are there ever actually situations where the answer is "no, I don't want to update the cache"?

I can think of 2 situations:
1. User is offline
	Updating cache doesn't make sense in this case. It will inevitably result in failure which might if not gracefully handled, prevent users from getting even the potentially stale page rendered.
2. User is asking for _classics_ like e.g. `tar`
	Sometimes users simply don't care about new additions to those pages but just want to get things done as [quickly as possible](https://xkcd.com/1168/). It feels wrong to me to force them to wait for latest and greatest changes if they only want to extract some tarball which was covered inside that page from 0-day.

agnivade:
I think a decent compromise might be to add a config flag for this, which can be overridden with a cmd line flag. IMO, asking the user for a yes/no confirmation dialog is not a good UX. It may make sense when something destructive is happening, like deleting a DB or a file. But not if some stale cache is going to be updated. This can be pre-defined behavior which can be overridden if the user wants.

agnivade:
Closing due to inactivity. It seems like the discussion has some to a standstill. Happy to reopen this if the discussion sparks up again.

Thanks everyone for the discussion.

PR #292, opened 2020-04-21
Reduce dependency and number of calls for config.get

OiYouYeahYou:
## Description

Now at runtime only `bin/tldr` depends on `config.js`. Which then passes
the result to the rest of the app. Before this, the function was called
multiple times, resulting in excessive fs calls just to get one value
out. This will hopefully make the tool faster and easier to write tests.

Update "should use randomly created temp folder" to become "should
created a temp folder" in `cache.spec.js`. And add "should generate
randomly salted temp directory paths" in `config.spec.js`. As config now
has the responsibility of defining the temp dir location.

Update `index.spec.js` and `search.spec.js` to pass config to methods
that now depend on it.

**Before:**

before

**After:**

after

<small>(please excuse the `bin/tldr.js` madge wouldn't reconginse the extensionless file as Javascript)</small>

### Why

I want to make some further contributions. And I felt it would make it easier
to contribute if the config was more central in defining certain variables.
While making it's interface dependants require injection instead of importing
it throughout the application. For example this approach will enable writing
future tests that can avoid depending on a globally installed version
`~/.tldr` directory

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/lib/remote.js
+++ b/lib/remote.js
@@ -1,19 +1,18 @@
 'use strict';
 
 const unzip = require('node-unzip-2');
-const config  = require('./config');
 
 let request = require('request');
 
 // Downloads the zip file from github and extracts it to folder
-exports.download = (path) => {
-  let url = config.get().repository;
+exports.download = (config, path) => {
+  let url = config.repository;
 
   // Creating the extractor
   let extractor = unzip.Extract({ path });
 
   // Setting the proxy if set by config
-  if (config.get().proxy) {
+  if (config.proxy) {
     request = request.defaults({ proxy: config.proxy });
   }
 
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -1,19 +1,17 @@
 'use strict';
 
 const fs = require('fs-extra');
-const os = require('os');
 const path = require('path');
 const remote = require('./remote');
 const platform = require('./platform');
 const index = require('./index');
-const utils = require('./utils');
 
 class Cache {
   constructor(config) {
     // TODO: replace this with a private field when it reaches enough maturity
     // https://github.com/tc39/proposal-class-fields#private-fields
     this.config = config;
-    this.cacheFolder = path.join(config.cache, 'cache');
+    this.cacheFolder = config.cacheFolder;
   }
 
   lastUpdated() {
@@ -23,7 +21,7 @@ class Cache {
   getPage(page) {
     let preferredPlatform = platform.getPreferredPlatformFolder(this.config);
     const preferredLanguage = process.env.LANG || 'en';
-    return index.findPage(page, preferredPlatform, preferredLanguage)
+    return index.findPage(this.config, page, preferredPlatform, preferredLanguage)
       .then((folder) => {
         if (!folder) {
           return;
@@ -42,7 +40,7 @@ class Cache {
 
   update() {
     // Temporary folder path: /tmp/tldr/{randomName}
-    const tempFolder = path.join(os.tmpdir(), 'tldr', utils.uniqueId());
+    const tempFolder = this.config.tempFolder;
 
     // Downloading fresh copy
     return Promise.all([
@@ -52,7 +50,7 @@ class Cache {
     ])
       .then(() => {
         // Download and extract cache data to temporary folder
-        return remote.download(tempFolder);
+        return remote.download(this.config, tempFolder);
       })
       .then(() => {
         // Copy data to cache folder
@@ -62,7 +60,7 @@ class Cache {
         return Promise.all([
           // Remove temporary folder
           fs.remove(tempFolder),
-          index.rebuildPagesIndex()
+          index.rebuildPagesIndex(this.config)
         ]);
       })
       // eslint-disable-next-line no-unused-vars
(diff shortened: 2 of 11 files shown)


lib/remote.js
```diff
@@ -1,19 +1,18 @@
...
+exports.download = (config, path) => {
+  let url = config.repository;
 
   // Creating the extractor
   let extractor = unzip.Extract({ path });
 
   // Setting the proxy if set by config
-  if (config.get().proxy) {
+  if (config.proxy) {
     request = request.defaults({ proxy: config.proxy });
```
OiYouYeahYou, line 16: Possibly fixed a bug here, as it was looking for proxy at the module scope

agnivade:
Thanks for this ! Yes the config global variables were a source of pain.

I will find some time to review this, and then we can get to merging.

OiYouYeahYou:
@sbrl I would agree with that in some cases, but not this one. The `config.get()` function, while simple, isn't trivial in constructing the config interface. So it continuing to be a function is more convenient than a flat `module.exports = {/*all the configs belong to us*/}`.

@agnivade I look forward to it, thank you

vladimyr:
Now I'm in a tight spot... 😩 

@OiYouYeahYou I appreciate your effort and I very much like your execution-style with dependency graphs and all the stuff! But...

> Now at runtime only `bin/tldr` depends on `config.js`. Which then passes
> the result to the rest of the app.

That is called dependency injection and I find it a really hard thing to sell in the nodejs world. Also if you would go full-on DI you should probably need to convert all those modules to factories instead of explicitly passing it to every single method. It is pretty much a complete rewrite and the only thing that is missing is some sort of received benefit.

Here is where another part of your statement comes into play:

> Before this, the function was called
> multiple times, resulting in excessive fs calls just to get one value
> out.

That is very much true and very much flawed engineering. Kudos for spotting and trying to fix it. However, you took a canon to kill a fly which @sbrl clearly pointed out:

> Another way to solve this issue would be to export the result of the `.get()` function, instead of the `get()` function itself. This would work, because Node.js evaluates a module once - and caches the result for subsequent imports.

And to add on it, yet another way is to add memoize `.get()` function so it assembles config only on the first invocation and later on returns an internal cached result.

> @sbrl I would agree with that in some cases, but not this one. The `config.get()` function, while simple, isn't trivial in constructing the config interface. So it continuing to be a function is more convenient than a flat `module.exports = {/*all the configs belong to us*/}`.

I don't get what you are talking about. If you need a block - write a block, it is that simple. You could always write it like an IIFE:
```js
module.exports = (function () {
  const config = doTheMagic();
  return config;
}());
```

To conclude:
1. It is always better to start with an issue and discussion
2. This PR in its current state is no-go for me. I would gladly accept memoization of `config.get` but a major overhaul of tool's internals - sorry - but no. It is risky, makes codebase messy, and all that just to fix repeated `fs.readFileSync` invocations which could be fixed way easier...

I don't want to discourage you in any way and there is also an open possibility for @agnivade to override me. My point here is because you can do something it doesn't mean you should do it, let's discuss it first, please.

agnivade:
Thanks for the analysis @vladimyr - I haven't looked at the code yet. But I'll keep your comments in mind.

agnivade:
So I looked into it. And I agree with @vladimyr's analysis that the current approach makes the code too messy.

If we look back into the history a bit, the `config.get()` call was indeed cached. But it was a global variable. I removed that here: https://github.com/tldr-pages/tldr-node-client/commit/18d250f59e19e56c4ebd23d0390751e0cdd670c7.

After that, what I did was refactored everything to be classes and objects, so that the config is initialized once in bin/tldr and passed down the tree to dependent objects. You can see that here: https://github.com/tldr-pages/tldr-node-client/commit/5c6b825556a580a0807ceb4741797a5f73f4af64.

However, I just did that for `cache` and `tldr`, and didn't get time to revamp the search and index functionalities too. I believe this is the factory approach that @vladimyr is referring to here:

> Also if you would go full-on DI you should probably need to convert all those modules to factories instead of explicitly passing it to every single method. 

In short, I would suggest we follow that approach and convert other components to objects too. Mainly because the codebase is already trying to move towards a class based approach where global API surface is minimized and every object is initialized with a constructor, and has internal state which is not leaked outside. This maintains everything you want - 
- fewer syscall to the system.
- easier tests.
- simpler state management.

Yet, this does that in a way that is synergistic with the rest of the codebase and is the right way forward IMO. Doing it in your way passes the config object to _each and every_ function which is a lot of effort when, I believe, we can get to this state with the same amount of effort.

With that being said, I really appreciate the time and effort you have put into this. But as pointed out earlier, had you opened an issue trying to discuss the problem and the approach you are taking, it would have been much easier to see this coming.

But now, I think you are already a lot familiar with the codebase. Would you be open to a new PR to start doing the refactor one component at a time and slowly migrate the `search` and `index` components ? If not, that's totally fine too. And I really appreciate your effort in trying to clean the codebase. Thanks again :smile:

agnivade:
@OiYouYeahYou - any thoughts on this ?

agnivade:
Closing due to inactivity. Please let me know whenever you are ready, and we can revisit this. Thanks again for your time :)

PR #293, opened 2020-04-21
Add docker testing environment

OiYouYeahYou:
## Description
This adds a simple Dockerfile and script that allows targetting specific
node versions and inuslates the host machine from from changes. Meaning
you can get feedback on changes for older versions without swapping out
node versions.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/docker-test
+++ b/docker-test
@@ -1,9 +1,17 @@
-# This script build and runs a dokcer enviroment for testing
+#!/usr/bin/env sh
+
+# This script build and runs a docker environment for testing
 # 
 # ARGS: 
 #  1: Node JS version
-#  2: Alternative npm registry url
+#  2: Alternative npm registry URL
 #
+
+which "$docker" &> /dev/null
+if $?; then
+	echo "\n\nDocker is not installed\n\n"
+	exit 127
+fi
 
 set -ex
 NODE_VERSION="${1:-6.12.0}"
--- a/docker.sh
+++ b/docker.sh
@@ -1,0 +1,23 @@
+# This script build and runs a dokcer enviroment for testing
+# 
+# ARGS: 
+#  1: Node JS version
+#  2: Alternative npm registry url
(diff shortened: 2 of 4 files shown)


OiYouYeahYou:
To add: I used `v6.12.0` because that was the version in `package.json`. The version was release in 2017, I'm open to changing the version

vladimyr:
Thanks for your contribution but how about using https://github.com/ehmicky/nve/blob/master/README.md#examples-multiple-versions instead?

OiYouYeahYou:
@vladimyr I like the look of that and might get some use out of it, thank you. Though I still would like to contribute these changes

sbrl:
The [official Node.js website](https://nodejs.org/en/) says that v12.x is the latest stable version - perhaps we could use that instead?

vladimyr:
@OiYouYeahYou Let me explain my reasoning. This is pretty much a dev helper and while you find it useful (and believe me, I do too) it is not about that but rather does add it to the codebase justifies increased maintenance cost. Sure it works now but someone needs to be able to go and update that dockerfile and shell script if needed and folks around are mainly js devs.

Leaving that aside, before you even try to win me over this one you need to do a few things:
1. add shebang to your shell script
2. pass it through [`shellcheck`](https://www.shellcheck.net/)
3. fix spelling error in the description
4. make sure to check if `docker` is installed before you start doing things
5. add it to `package.json` scripts

agnivade:
What is the problem that this is trying to solve ?

Is this about making it easy for devs to swap out Node versions ? I thought `nvm` was a commonly used tool for that ? Is this a common practice in other projects too ?

agnivade:
Hi @OiYouYeahYou - I am going to close this one due to inactivity. However, please feel free to comment whenever you are ready to continue the discussion and we can reopen. Thank you.

PR #294, opened 2020-04-24
Tidy up tests

OiYouYeahYou:
## Description

* Add nyc coverage tool
* Remove `env-test`: it's a single line for the cost of a module
* Set NODE_ENV in custom set up file
* Stub console for each test individually globally
* Plus: a `sinon.restore()` in global afterEach prevents cross test leaks
* Add most basic test for `Tldr` class, just to trigger coverage
* Remove surplus mocha args:
        * It always looks in test dir
        * `mocha.opts` provides a more consistent approach
* Ignore coverage files now created by nyc

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/test/mocha.js
+++ b/test/mocha.js
@@ -1,4 +1,4 @@
-'use strict'
+'use strict';
 
 process.env.NODE_ENV = 'test';
 
--- a/test/mocha.js
+++ b/test/mocha.js
@@ -1,0 +1,11 @@
+process.env.NODE_ENV = 'test';
+
+const sinon = require('sinon');
+
+beforeEach(() => {
+  sinon.stub(console);
+});
+
+afterEach(() => {
+  sinon.restore();
+});
--- a/test/mocha.js
+++ b/test/mocha.js
@@ -1,3 +1,5 @@
+'use strict'
+
 process.env.NODE_ENV = 'test';
 
 const sinon = require('sinon');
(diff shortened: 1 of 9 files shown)


test/mocha.js
```diff
@@ -0,0 +1,11 @@
+process.env.NODE_ENV = 'test';
```
vladimyr, line 1: Move this to `mocha.opts` as `--env` flag cause that is where people would typically look for it. Add `use strict` pragma on the top of the file.

OiYouYeahYou: Could you give me the syntax of a `--env` flag with mocha? I'm looking for it's documentation and have thrown some stuff at the wall, but nothing seems to work.

test/mocha.opts
```diff
@@ -1,3 +1,3 @@
 --reporter spec
```
agnivade, line 1: While we are here, can we move from mocha.opts to a config file ? Looks like it is deprecated and will be removed in future: https://mochajs.org/#mochaopts. `.mocharc.js` looks fine by me.

test/mocha.js
agnivade: I don't think there _is_ an `--env` flag: https://github.com/mochajs/mocha/issues/185. I am good with this or setting it via a variable in the npm invocation string.

PR #295, opened 2020-04-25
Remove sanitization

agnivade:
## Description

The marked library deprecated its sanitizer option
and recommends to use a separate library to do all the sanitization.

Since we are a CLI tool focused on showing in the terminal, there's no particular need to sanitize HTML. Let's just remove the sanitize option altogether.

## Checklist

- [x] Code compiles correctly
~- [ ] Created tests, if possible~


agnivade:
@vladimyr - Any thoughts on this ?

sbrl:
I'm probably just ignorant, but where's the HTML coming from that needs sanitizing?

If it's  the output of a markdown parser, then doesn't the markdown parser have an option to make the output safe?

agnivade:
> If it's the output of a markdown parser, then doesn't the markdown parser have an option to make the output safe?

That's what they have deprecated. The library thought they were taking on too much responsibility. Thus they have delegated that to any 3rd party library.

sbrl:
I see. Is there an option to disable HTML in the markdown parser then? tldr pages don't contain HTML, so maybe that could help make it secure?

agnivade:
Well, a markdown library is supposed to generate HTML, so I don't think there is an option. https://marked.js.org/#/USING_ADVANCED.md#options.

The thing is that we are using the markdown parser to hook into various components to fill our page struct. We aren't even rendering the HTML.

Hence my question of whether it's even worth trying to sanitize something we aren't even using.

sbrl:
Ah, I  don't think that's quite what I meant - but it seems that it doesn't have an option for what I was thinking of anyway. Some markdown parsers (such as Parsedown in PHP) support an option that allows you to control whether the markdown parser will allow HTML in the original markdown source. If such an option existed if I understand correctly setting it to false would solve the issue. 

Ah, I see. In that case I  think there's limited point in santizing at all, especially  since we aren't rendering the HTML in a browser.

I wonder if you could make an abstract syntax tree out of the markdown directly and use that instead? Then you wouldn't even finish the parsing process - improving performance - though that would likely be a lot of work for limited benefit.

agnivade:
> I wonder if you could make an abstract syntax tree out of the markdown directly and use that instead? Then you wouldn't even finish the parsing process - improving performance - though that would likely be a lot of work for limited benefit.

I agree. I'll just remove the sanitization. If somebody else wants to improve performance, they are more than welcome.

PR #298, opened 2020-05-12
Replace tab with 2 spaces

zdroid:

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Meanwhile, you can do one of the following to fix this issue:
 ```bash
 tldr_path="$(which tldr)"
 function tldr() {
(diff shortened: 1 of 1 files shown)


vladimyr:
That moment in life when failing functional test prevents you from changing the indentation style in the project's readme. 🏆

sbrl:
Not sure I understand what's going on here. Personally I'd prefer tabs, but what does the repo style guide (if there is one) say?

Either way I'm not sure this actually warrants a PR.

zdroid:
@sbrl Tabs should have been automatically replaced by `.editorconfig`[https://github.com/tldr-pages/tldr-node-client/blob/master/.editorconfig]... but they weren't.

PR #300, opened 2020-05-27
Remove snap install

### LINKED ISSUE #289: Stable snap package is broken
Running Ubuntu 18.04, but it's snap so it shouldn't matter.

The stable snap package is 3.1.0 from 2017, and it doesn't work. When using tldr the first time, it says "page not found, updating cache", and then just eats CPU forever. Uninstalling the snap then installing the --edge snap (3.3.2), tldr updates the cache properly and works.

Considering how old the snap is, it's safe to presume it's abandoned. I think if users are meant to use the edge channel only, then you should unpublish the stable package from the snap store, to avoid having people waste their time troubleshooting.

agnivade:
It is broken and it does not work.

Fixes #289

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,22 +18,6 @@ A `Node.js` based command-line client for [tldr](https://github.com/tldr-pages/t
 npm install -g tldr
 ```
 
-## Install from the snap store
-
-In any of the [supported Linux distros](https://snapcraft.io/docs/core/install):
-
-```bash
-sudo snap install tldr
-```
-
-If you want to help testing the latest changes, and get the newer features earlier, you can install the snap from the edge channel:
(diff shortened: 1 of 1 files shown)


agnivade:
Done !

sbrl:
Hrm, for some reason it's not giving me a button to force-merge it. Let me try on the command-line.

sbrl:
Ah, it's because it's a protected branch. One sec while I disable that for a brief moment.

sbrl:
Done.

It looks like we couldn't merge it because the "enforce rules for administrators option is turned on. I've left it like that for now - I just renamed the branch pattern for a sec.

PR #303, opened 2020-06-09
Migrate build to GitHub actions

MasterOdin:
## Description

This migrates the test toolchain from Travis-CI to GitHub actions. They should operate relatively the same, with the exception that `latest` is not a valid node version ([yet](https://github.com/actions/setup-node/issues/26)). I did add 13.x and 14.x though to be explicitly tested, and dropped 11.x as that's EOL.

## Checklist

- [x] Code compiles correctly


agnivade:
This is great stuff. While we are on this, should we just drop the windows check ? It's flaky and doesn't work half of the time.

MasterOdin:
I can add windows to be tested within Github actions as well as the other OSes, and can then also delete the appveyor.yml config (I missed that it was there). The main thing would be probably to add in a .gitattributes file to normalize line endings by default as no point having a style guide that mandates line endings but then allow people to check out the repo on windows using CRLF (and so the lint will always fail).

agnivade:
Makes sense. Thanks for this @MasterOdin !

vladimyr:
@MasterOdin This is great!

#304 landed so feel free to proceed with Appveyor to Github actions migration.

Just two small things to add on your todo list:
- [ ] Make sure that action invocations are autocancelled on subsequent branch pushes. I'm not super experienced with Github actions but I believe there are some _addons_ for that kind of thing.
- [ ] Update readme badges using correct badge from https://badgen.net

Once again, thank you for your effort 👍

owenvoke:
Looks like cancelling previous builds when new stuff is pushed isn't natively supported, however there is a GitHub Action for this:
https://github.com/marketplace/actions/cancel-workflow-action

MasterOdin:
Enabling windows at this moment is actually not possible given that appveyor has been failing for 7 months. Fixing the windows tests should probably be a separate PR.

> Make sure that action invocations are autocancelled on subsequent branch pushes. I'm not super experienced with Github actions but I believe there are some addons for that kind of thing.

> Looks like cancelling previous builds when new stuff is pushed isn't natively supported, however there is a GitHub Action for this:
> https://github.com/marketplace/actions/cancel-workflow-action

It should be noted this only supports pushes to master, not pushes to PRs. I can add in that step if that's desired, but I would note that running the current suite on 4 versions on 2 OSes takes ~3 minutes per commit. It's an unfortunate limitation of GH actions vs travis.

owenvoke:
Interesting, with dependencies cached I've found GitHub Actions to be much faster than Travis.

vladimyr:
> It should be noted this only supports pushes to master, not pushes to PRs.

I haven't found this in their docs? Where does that limitation come from?

README.md
```diff
@@ -1,8 +1,7 @@
 # tldr-node-client
 
 [![NPM version][npm-image]][npm-url]
-[![Travis CI Build Status][travis-image]][travis-url]
-[![AppVeyor CI Build status][appveyor-image]][appveyor-url]
+[![Travis CI Build Status][gh-actions-image]][gh-actions-url]
```
vladimyr, line 4: Update the label, please 😉

MasterOdin:
I've updated the badges in the README to use https://shields.io since that's where the travis / appveyor / npm badges were from.



> > It should be noted this only supports pushes to master, not pushes to PRs.
> 
> I haven't found this in their docs? Where does that limitation come from?

It's a limitation in that to cancel a running action, you must go through the actions API. This API is restricted (as would be expected) to only people with read / write access to the repository. Forked repositories do not have that permission (as expected), and so PRs from forked repositories (e.g. like this one) cannot cancel currently running actions then. You could probably implement some amount of logic to determine whether or not a PR is from a forked repository or not if you really wanted cancelling in as many cases as possible, but I'm not sure the full details on this.

README.md
MasterOdin: whoops. done.

vladimyr:
> Enabling windows at this moment is actually not possible given that appveyor has been failing for 7 months. Fixing the windows tests should probably be a separate PR.

Ok, then just comment out AppVeyor badge in the readme and we'll take care of it later.

> I've updated the badges in the README to use https://shields.io since that's where the travis / appveyor / npm badges were from.

Yeah, no worries, even though I preferer badgen being consistent is a more important 👍

vladimyr:
Regarding autocancelling, thank you for the detailed explanation, much appreciated! Let's do the best we can ATM. Put autocancel as part of the workflow. It won't do any harm and might save us some minutes for internal PRs at least.

vladimyr:
@MasterOdin Just please bring back AppVeyor config and hide the badge. So one can easily inspect the logs there as part of some future _fix the windows build_ effort.

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest] # todo: windows-latest
+        node-version: [10.x, 12.x, 13.x, 14.x]
+
+    steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.4.0
+      if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.id == github.event.pull_request.base.repo.id }}
```
vladimyr, line 21: Wouldn't this prevent executing it for subsequent non-merge commits to the master branch? 🤔

MasterOdin: Please let me know what events you want to cancel.

vladimyr: I would argue we should cancel previous workflow runs:
1. when new commit lands on the PR branch
2. when new commit lands on the master branch

What do you think? Does that make sense?

MasterOdin: Yeah, this line covers that (though it will cancel commits on any branch, and not just master). Merging a PR to master counts as a push to master (which is picked up as `github.event_name != 'pull_request'`) as well as just regular pushes to master.

vladimyr: @MasterOdin Sorry for causing confusion I've somehow managed to interpret it inverted inside my head 🤦 
Yup, this seems ok 👍

agnivade:
So travis appears to be stuck on CI. Is that expected ?

owenvoke:
Probably because the Travis config has been removed, however it is required as part of the [`master` branch protection rules](https://github.com/tldr-pages/tldr-node-client/settings/branch_protection_rules/88868).

agnivade:
Amazing, let's give this a go.

owenvoke:
Well, looks like it's working 🎉 Nice work @MasterOdin!

PR #304, opened 2020-06-09
 Add gitattributes to force lf eol in git

MasterOdin:
## Description

This makes it so that regardless of a user's settings in git, always use LF line endings. If this is not set, then a user on Windows, when they do git clone, might have a repo where all files have the CRLF line ending, which will cause the linter to fail, which can be confusing to an end user.

Once this is merged, I will update #303 to include a windows build target in the github action (as right now, it would fail without normalizing line endings, see https://github.com/MasterOdin/tldr-node-client/runs/754519436?check_suite_focus=true.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


owenvoke:
Is the `eol=lf` required? I always thought that was default if `text=auto` is enabled.

MasterOdin:
Yes, though my personal style is to prefer explicit with anything in the .gitattributes file as in my experience, the git docs are a bit hard to read for a novice. I'm happy to remove it if desired.

vladimyr:
> Yes, though my personal style is to prefer explicit with anything in the .gitattributes file as in my experience, the git docs are a bit hard to read for a novice. I'm happy to remove it if desired.

I think we are fine keeping things explicit.

@MasterOdin Thank you for your contribution!

PR #308, opened 2020-06-10
fix running tests under windows

### LINKED ISSUE #305: Fix failing windows build
Looking at appveyor (and github actions), it looks like the windows builds have been failing the test suite for ~7 months or so, where it broke with #270.

MasterOdin:
## Description

The tests used a linux path separator for referencing things on the filesystem, except that then uses path.sep under the hood for path splitting, which is / on *unix and \\ on windows. Instead of hard-coding the separator, we can use path.join instead which will give us the right seperator.

Closes #305 

Testing on windows is activated under github actions.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/test/index.spec.js
+++ b/test/index.spec.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const fs = require('fs-extra');
+const path = require('path');
 const index = require('../lib/index');
 const utils = require('../lib/utils');
 const sinon = require('sinon');
@@ -25,7 +26,9 @@ const pages = [
   '/pages/sunos/dd.md',
   '/pages/sunos/du.md',
   '/pages/sunos/svcs.md'
-];
+].map((x) => {
+  return path.join(x);
+});
 
 describe('Index building', () => {
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


sbrl:
Isn't [Node 13.x out of support](https://nodejs.org/en/about/releases/)?

test/index.spec.js
```diff
@@ -25,7 +26,9 @@ const pages = [
   '/pages/sunos/dd.md',
   '/pages/sunos/du.md',
   '/pages/sunos/svcs.md'
-];
+].map((x) => {
+  return path.join(x);
+});
```
agnivade, line 31: This is a bit strange. Doing a `path.join` on a single string makes no difference at all. What is the objective here?

MasterOdin: On windows, it converts the paths to use windows delimiters, so that for example, `/pages/sunos/svcs.md` becomes `\\pages\sunos\\svcs.md`. Without doing this, the tests in this suite fail as this is then used directly as mock data from methods not being tested (`utils.walk`).

agnivade: That's interesting. So even without passing individual parts of a file path, `path.join` splits them and reassembles them? If that is the case, then why are we passing individual strings in other cases? We can just wrap the string with a path.join.

MasterOdin: I think if anything, it would make sense to break each part of this array up into pieces, such that it's:

```js
const pages = [
  ['/index.json'],
  ['/pages', 'linux', 'apk.md'],
  ['/pages.zh', 'linux', 'apk.md'],
 // the rest of them
]
```

to avoid the confusion on what this does versus utilizing this trick elsewhere in the file.

agnivade: That makes sense to me. Let's do that.

agnivade: Hi @MasterOdin - just wanted to check on this.

MasterOdin:
Recreated PR as #319 as this one has gotten broken and uneditable.

Issue #256, opened 2019-05-19
Please catch error when updating cache


linonetwo:
## Expected behaviour

No "internal error" thrown, instead, a catchable Error thrown.

## Actual behaviour
Tell us what happens instead.

Randomly throwing one of the following errors when Wifi is off:

```

A JavaScript error occurred in the main process
Uncaught Exception:
Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:182:27)

A JavaScript error occurred in the main process
Uncaught Exception:
Error: socket hang up
    at createHangUpError (_http_client.js:334:17)
    at Socket.socketOnEnd (_http_client.js:435:23)
    at Socket.emit (events.js:199:15)
    at endReadableNT (_stream_readable.js:1141:12)
    at processTicksAndRejections (internal/process/task_queues.js:81:17)
```

And I can't simply catch this error.

## Environment
- Operating system - OSX 10.14.4 (18E226)
- NodeJS in electron:

Electron Version: 5.0.1
Node Version: 12.0.0
Chrome Version: 73.0.3683.121

agnivade:
@linonetwo - Can you please check with tldr latest version 3.3.6 if the issue has been fixed? Thanks!

linonetwo:
Wow, glad to see this can get a fix, I was using this for https://github.com/linonetwo/zazu-tldr , but it is not maintained and not used recently...

I will revisit this in 2021 after I finish my other opensource projects, thank you!

PR #316, opened 2020-10-16
Address Race Conditions and Better Error Handling to fix Issue #256

mearns:
## Description

Better error handling to address what I think issue #256 is about. There were also a couple of antipatterns being used and improper use of promises that lead to race conditions:
    
1. Using `process.exit()`  directly is usually not advisable as it prevents things from shutting down nicely. It also means if someone is trying to use this as a library, which I think may be the case for issue #256, you will exit from their code as well (i.e., error handling becomes impossible).

2. Functions that caught errors were printing them directly to STDERR rather than rethrowing them or throwing a new error. Again, if someone is trying to use it as a library as in the issues, this will make it impossible for them to do error handling.

3. There were some places where errors were being caught and printed to STDERR, but not causing the the process to exit with a non-zero exit code. It's always advisable to exit with a non-zero exit code when there is a failure.
    
4. There were some places where promises were not being waited for, leading to race conditions.

## Checklist

- [X] Code compiles correctly
- [X] All tests passing (`npm run test:all`)

### DIFF
--- a/lib/tldr.js
+++ b/lib/tldr.js
@@ -11,7 +11,7 @@ const messages = require('./messages');
 const parser = require('./parser');
 const render = require('./render');
 const index = require('./index');
-const exit = process.exit;
+
 
 class Tldr {
   constructor(config) {
@@ -23,34 +23,33 @@ class Tldr {
 
   list(singleColumn) {
     let os = platform.getPreferredPlatformFolder(this.config);
-    index.commandsFor(os)
+    return index.commandsFor(os)
       .then((commands) => {
-        this.printPages(commands, singleColumn);
+        return this.printPages(commands, singleColumn);
       });
   }
 
   listAll(singleColumn) {
-    index.commands()
+    return index.commands()
       .then((commands) => {
-        this.printPages(commands, singleColumn);
+        return this.printPages(commands, singleColumn);
       });
   }
 
   get(commands, options) {
-    this.printBestPage(commands.join('-'), options);
+    return this.printBestPage(commands.join('-'), options);
   }
 
   random(options) {
     let os = platform.getPreferredPlatformFolder(this.config);
-    index.commandsFor(os)
+    return index.commandsFor(os)
       .then((pages) => {
         if (pages.length === 0) {
-          console.error(messages.emptyCache());
-          exit(1);
+          throw new Error(messages.emptyCache());
         }
         let page = sample(pages);
         console.log('PAGE', page);
-        this.printBestPage(page, options);
+        return this.printBestPage(page, options);
       })
       .catch((err) => {
         console.error(err);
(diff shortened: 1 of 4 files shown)


agnivade:
Thanks @mearns ! The PR seems to have the older commit which got merged. Let's remove that.

mearns:
Sure thing @agnivade. Except...I'm not sure how to =/. The only way I can think of is to close this PR and create a new branch from master. Sound ok?

agnivade:
You can do a `git rebase -i HEAD~2` and then remove the earlier commit. And then do a `git push -f origin issues/256`. That should remove the commit and for a force push and update the PR.

agnivade:
Thanks! I'd like @vladimyr to take a look.

mearns:
Thanks for the git-fu, I wasn't sure if rebase was appropriate after it had been pushed, but it looks like it worked well =)

agnivade:
Apologies for the delay here @mearns. I will get someone to review this.

lib/tldr.js
```diff
@@ -23,34 +23,33 @@ class Tldr {
 
   list(singleColumn) {
     let os = platform.getPreferredPlatformFolder(this.config);
-    index.commandsFor(os)
+    return index.commandsFor(os)
```
sbrl, line 26: This could potentially be refactored to use async / await to reduce visual complexity

mearns: Sure. I didn't see async/await in there so I didn't want to introduce it, but I do prefer using it so if you'd prefer that syntax I can update it.

agnivade: Up to you @mearns. Since you have already spent your time on it, I don't want to redo it again.

lib/tldr.js
```diff
@@ -59,138 +58,107 @@ class Tldr {
...
+          })
+            .then(() => {
+              return spinningPromise('Creating index...', () => {
+                return search.createIndex();
+              });
+            })
+            .then(() => {
+              // And then, try to check in cache again
+              return this.cache.getPage(command);
+            });
```
agnivade, line 141: Let's unindent these to match the closing branch of the previous block.

agnivade (commented):
Just one small comment. Looks good otherwise.

lib/tldr.js
mearns: Looks like the projects current configuration doesn't support async/await syntax; it comes up with linting errors. I tried a few changes to address it, including updating to eslint version 7 and setting`parserOptions.ecmaVersion` to 7, but neither of these worked.

More importantly, these keywords [aren't supported before Node v7.10](https://node.green/#ES2017-features-async-functions), so to add them would require dropping support for older versions of Node (currently specified in package.json:engines as `>=6.12.0`), or else adding a transpiler.

In light of this, I propose that adding async/await is out of scope for this change and is a decision that the project owners need to make. If the decision is made to add them, I'd be happy to assist with that, but I propose it should be in a separate change set (a dedicated PR and release).

lib/tldr.js
mearns: If I understood the request correctly, eslint complains about that. I deleted two leading spaces from lines 133 through 141 (inclusive), so it looks like this:

```javascript
        if (!content) {
          return spinningPromise('Page not found. Updating cache...', () => {
            return this.cache.update();
          })
          .then(() => {
            return spinningPromise('Creating index...', () => {
              return search.createIndex();
            });
          })
          .then(() => {
            // And then, try to check in cache again
            return this.cache.getPage(command);
          });
        }
        return content;
      })
```

This results in 8 eslint errors due to violating the "indent" rule.

agnivade: Well, that's a bit weird. But thanks for checking. :+1:

lib/tldr.js
agnivade: Let's go with the existing code for now.

agnivade (approved):
Thanks!

PR #319, opened 2020-11-22
fix running tests under windows

### LINKED ISSUE #305: Fix failing windows build

MasterOdin:
Recreating #308 as my fork got messed up there. I've made the suggested changes from that PR here.

## Description

The tests used a linux path separator for referencing things on the filesystem, except that then uses path.sep under the hood for path splitting, which is / on *unix and \\ on windows. Instead of hard-coding the separator, we can use path.join instead which will give us the right seperator.

Closes #305 

Testing on windows is activated under github actions.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest] # todo: windows-latest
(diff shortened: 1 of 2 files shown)


agnivade (approved):
Thank you!

PR #337, opened 2021-03-26
change `-o/--os` to `-p/--platform`

### LINKED ISSUE #336: `-o` and `--os` is not compliant with the client spec
According to the [client spec](https://github.com/tldr-pages/tldr/blob/master/CLIENT-SPECIFICATION.md#arguments), the arguments for platform-specific commands should be `-p` and `--platform`. However, here it is `-o` and `--os`.

Reference: https://github.com/tldr-pages/tldr/pull/5508#issuecomment-808466616

marchersimon:
Closes #336


marchersimon:
Any idea what's wrong with the failed test here?

bl-ue:
@marchersimon don't edit this PR yet—I'm making some changes.

bl-ue:
Closing→reopening the PR can frequently fix the checks, not the case here.  I think you may have broken it, let me see.

bl-ue:
The tests should pass now.  FYI @marchersimon when you change the flags in the code ([here](https://github.com/tldr-pages/tldr-node-client/pull/337/files#diff-3acbcdf8e402b3188bd91d12dbe9f21a90e627f1eb2cce88106d7e455405f0e3R23)) you also need to update references to that flag, e.g. `program.os` ([here](https://github.com/tldr-pages/tldr-node-client/pull/337/files#diff-3acbcdf8e402b3188bd91d12dbe9f21a90e627f1eb2cce88106d7e455405f0e3L64)) corresponds directly to the string values passed to `program.option`, so it needs to be changed to `program.platform`.  All of that and more I've done in the past two commits 👍🏻

marchersimon:
Oh, sorry, I just blindly replaced `-o` and `--os` without checking if they occur somewhere else. Thanks for the fix

bl-ue:
There we go, the checks are all green.

bl-ue (approved):
Funny that the most official of all the official clients doesn't even comply with the spec (in a big way!) 😄

bin/tldr
```diff
@@ -4,7 +4,7 @@ const program = require('commander');
 const pkg = require('../package');
 const Tldr = require('../lib/tldr');
 const config = require('../lib/config');
-const platform = require('../lib/platform');
+const platformUtils = require('../lib/platformUtils');
```
vladimyr, line 7:
```suggestion
const Platform = require('../lib/platform');
```
Makes diff smaller and addresses name clash in a fairly standard manner.

bin/tldr
```diff
@@ -20,11 +20,11 @@ program
   .option('-e, --random-example', 'Show a random example')
   .option('-f, --render [file]', 'Render a specific markdown [file]')
   .option('-m, --markdown', 'Output in markdown format')
-  .option('-o, --os [type]', 'Override the operating system [linux, osx, sunos, windows]')
-  .option('--linux', 'Override the operating system with Linux')
-  .option('--osx', 'Override the operating system with OSX')
-  .option('--sunos', 'Override the operating system with SunOS')
-  .option('--windows', 'Override the operating system with Windows')
+  .option('-p, --platform [type]', 'Override platform [linux, osx, sunos, windows]')
```
vladimyr, line 23:
```suggestion
  .option('-p, --platform [platform]', 'Override platform [linux, osx, sunos, windows]')
```

vladimyr (commented):
Looks good overall, just a few suggested changes 👍
@marchersimon Thanks for addressing it.

vladimyr:
> Closing→reopening the PR can frequently fix the checks, not the case here.  I think you may have broken it, let me see.  I'll also re-run the jobs.

OT Please don't do that to minimize notification noise. You can always manually rerun Github actions through the web interface.

bl-ue:
You're right; I realized that after I closed/reopened it 😂

bin/tldr
bl-ue: You think so? I mean to ask you in the extended commit message of 5e1b052af5b21f82b93e57bdc5caccf9203bddef

vladimyr: Well, I've seen it more than once in various node codebases but that is anecdotal evidence. Maybe going with
```js
const platforms = require('../lib/platforms');
```
strikes the right balance?

vladimyr (changes requested):

bl-ue:
@vladimyr this is all great stuff that you're pointing out, but as it doesn't actually improve the app from the users's perspective (i.e. performance, etc.) what do you think about ignoring all of these tweaks and saving them for another PR?  I plan to open a new PR within the next couple days full of improvements and I think these would—as they're not strictly related to the `os` → `platform` change—be better for that PR.  I have a lot more things in mind than you're pointing out here, and IMO it makes the diff harder while not giving the code the wash it deserves.  What do you think?

agnivade (changes requested):
Thanks for the PR. However, changing an option outright is a huge breaking change. Let's _add_ the `--platform` option which can be a synonym for `--os` so that existing users aren't affected.

bl-ue:
Yeah you're right @agnivade, funny that I didn't think about that.  What about also printing a deprecation message when `-o`/`--os` are used, stating that they will be removed in a future release or after a specific date e.g. Dec 31 2021?

agnivade:
Sounds reasonable. It can be removed in the next major release.

bin/tldr
bl-ue: I don't have strong opinions regarding either, but do you think `platformUtils` it not standard?

marchersimon:
Should I still handle suggestions or would you prefer to take over? Because I think I wouldn't be very helpful here.

bl-ue:
Oh, I forgot to hit send.  I'll take over, provided you don't mind.

marchersimon:
Sure, thank you

bin/tldr
vladimyr: `utils` and `helpers` are well-known wildcards conveying absolutely zero meaning so some folks consider naming things using them as an antipattern. In this case, it has been used as name-clash avoiding sufix so it isn't really a problem excerpts it hurts my eyes 🙈
But let's take a step back, what exactly mandates rename in the first place, I failed to spot that name clash we are trying to avoid here? 😕

vladimyr:
> @vladimyr this is all great stuff that you're pointing out, but as it doesn't actually improve the app from the users's perspective (i.e. performance, etc.) what do you think about ignoring all of these tweaks for this PR and saving them for another PR? I plan to open a new PR within the next couple days full of improvements I've noticed could be made and I think these would—as they're not strictly related to the `os` → `platform` change—be better for that PR.
> 
> I have a lot more things in mind than you're pointing out here, and IMO it makes the diff harder while not giving the code the wash it deserves. What do you think?

So here is what I think. You are right changes I proposed are transparent to the end-user and don't affect them in any possible way but they improve the state of a particular piece of codebase limited by boundaries of this patch/PR. I tend to follow Martin Fowler's school of thought there which can be summarised as - each time you touch some part of the codebase go and make it a bit better right away instead of scheduling grand refactors.

As per your second point, I fully agree that there are _countless_ things that need to get addressed in tldr's codebase but please make that an incremental process rather than one big rewrite.

This last point of yours about making the diff harder I can just say I explicitly limited myself to not suggest too many changes and to keep those on topic. I think I did a good job there but that's obviously subjective 🙃

vladimyr:
Also as @agnivade pointed out this is a breaking change so increasing of major version is a semver requirement.

For displaying deprecation message, we could use https://www.npmjs.com/package/depd#deprecating-property-access

bl-ue:
>For displaying deprecation message, we could use https://www.npmjs.com/package/depd#deprecating-property-access

Looks nice, I'll try it.  The message might be developer-oriented rather user-oriented, but we'll see.

>So here is what I think. ...
>...
>This last point of yours about making the diff harder I can just say I explicitly limited myself to not suggest too many changes and to keep those on topic. I think I did a good job there but that's obviously subjective 🙃

Good points, I agree with all of them.  I'll apply the suggestions and keep future cleanup PRs small and limited to a particular type of fix.

bin/tldr
bl-ue: I want to rename all internal references to `os` to `platform` where appropriate, but I couldn't easily in some places because of `const platform = require("..../platform.js")`, get it?

vladimyr: That's the same situation as requiring the path module and then having a variable named path. In most cases platform is really accessed as a property of the program object which makes it a non-issue. And if there are some shadowing instances they should be addressed by renaming the local variable instead, for example, `targetPlatform` or `selectedPlatform`.

bl-ue: Yeah, you're thinking the right way.  I guess I'm busy so I didn't think a lot about this.  I'll change the name back to `platform.js` 👍🏻

bl-ue: What cases like this?
```javascript
function validatePlatform(platform) {
  if (platform && !platformUtils.isSupported(platform)) {
    return 'Unsupported platform : ' + platform;
  }
  return null;
}
```
Personally I'd pick `p` as a parameter name, but that doesn't strike my memory as conventional JS (I'm not working on much js lately)

vladimyr: Well if you do validation then one thing is for sure, you are doing it on `input` 😉

vladimyr: > Yeah, you're thinking the right way. I guess I'm busy so I didn't think a lot about this. I'll change the name back to `platform.js` 👍🏻

Watching me doing things you would soon realize that most of the time it seems like I'm not thinking at all 😂

lib/index.js
```diff
@@ -60,11 +60,11 @@ function findPage(page, preferredPlatform, preferredLanguage) {
       } else if (targets.length > 0 && hasLang(targets, preferredLanguage)) {
         targetLanguage = preferredLanguage;
         targetPlatform = targets[0].platform;
-        console.log(`Command ${page} does not exist for the host platform. Displaying the page from ${targets[0].platform} platform`);
+        console.log(`Command ${page} does not exist for the host platform. Displaying the page from ${targetPlatform} platform`);
       } else if (targets.length > 0 && hasLang(targets, 'en')) {
         targetLanguage = 'en';
         targetPlatform = targets[0].platform;
-        console.log(`Command ${page} does not exist for the host platform. Displaying the page from ${targets[0].platform} platform`);
+        console.log(`Command ${page} does not exist for the host platform. Displaying the page from ${targetPlatorm} platform`);
```
vladimyr, line 67: And I made yet another spelling mistake 🤦

bl-ue:
```suggestion
        console.log(`Command ${page} does not exist for the host platform. Displaying the page from ${targetPlatform} platform`);
```

sbrl:
I'm sure there'll be plenty of other PRs to the Node.js client here to improve things in too lol

(did we update the Node.js client to automatically convert the page name to lowercase?)

bl-ue:
No, I don't think so!  Another spec compliance issue!  I will gladly take care of that ASAP once this PR is merged.

agnivade:
Is this ready for review? This still seems to replace `--os` by `--platform`.

marchersimon:
Oh, sorry, you're right. I just saw that the PR hasn't had activity for some time.

sbrl:
I think this PR is indeed ready for review, @agnivade

vladimyr:
> I think this PR is indeed ready for review, @agnivade

It still lacks deprecation notice and as @agnivade already said simply replaces `--os` by `--platform` thus introducing a breaking change. :warning: ie this needs more work/love...

MasterOdin:
Closing this in favor of #421

PR #338, opened 2021-03-29
Fix debug messages

### LINKED ISSUE #335: tldr is showing a debug message when a page was not found
When I run tldr on a command which could not be found, I get the following output:
```
✔ Page not found. Updating cache...
✔ Creating index...
Error: Page not found.
If you want to contribute it, feel free to send a pull request to: https://github.com/tldr-pages/tldr
    at /usr/lib/node_modules/tldr/lib/tldr.js:147:17
```
Does the last line have any purpose? It's always the same and not really useful.
(tldr version 3.3.7)

marchersimon:
Closes #335


bin/tldr
```diff
@@ -119,7 +120,11 @@ if (p === null) {
   process.exitCode = 1;
 } else {
   p.catch((err) => {
-    console.error(err);
+    if(err instanceof TldrError) {
```
bl-ue, line 123:
```suggestion
    if (err instanceof TldrError) {
```

bl-ue (changes requested):

bl-ue (approved):

vladimyr:
@marchersimon Please apply these changes: https://github.com/vladimyr/tldr-node-client/commit/51d280c

The easiest way to do that is:
```sh
# reset your branch to master
git reset --hard master
# download patch from Github
curl -sSL https://github.com/vladimyr/tldr-node-client/commit/51d280c.patch -o debug-message.patch
# create commit from patch
git am debug-message.patch
```

Then you'll need to force push your branch but don't forget to use `--force-with-lease` flag :warning: 
```sh
git push --force-with-lease
```

lib/errors.js
```diff
@@ -0,0 +1,49 @@
...
+module.exports = {
+  TldrError,
+  EmptyCacheError,
+  MissingPageError
+};
+
+function trim(strings, ...values) {
+  let output = values.reduce((acc, value, i) => {
+    return acc + strings[i] + value;
+  }, '') + strings[values.length];
```
bl-ue, line 47: @vladimyr is there a reason you didn't do this?
```suggestion
  let output = values.reduce((acc, value, i) => acc + strings[i] + value, '') + strings[values.length];
```

lib/errors.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  static isTldrError(err) {
+    return err instanceof this;
+  }
+}
+
+class EmptyCacheError extends TldrError {
+  constructor() {
+    super(trim`
```
bl-ue, line 16: Cool trick! ⚡

lib/errors.js
vladimyr: Unfortunately, ESLint wouldn't let me do it 🙁
https://github.com/tldr-pages/tldr-node-client/blob/7e49dda42a11dfc07b5547b1bc9a20bc334b8792/.eslintrc.json#L9

bl-ue: @MasterOdin / @agnivade do you have any preferences regarding this style?  If you don't I'd like to relax this rule.

vladimyr: My guess is it will effectively double down the LOC amount...

marchersimon:
Can this be merged?

lib/errors.js
bl-ue: Let's wait for a different time.

bl-ue (approved):
It works!

owenvoke (approved):

lib/errors.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+class TldrError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'TldrError';
+  }
+
+  static isTldrError(err) {
```
agnivade, line 9: Can we make this a global function rather than a static method?

lib/errors.js
```diff
@@ -0,0 +1,49 @@
...
+
+class EmptyCacheError extends TldrError {
+  constructor() {
+    super(trim`
+Local cache is empty
+Please run tldr --update
+    `);
+    this.name = 'EmptyCacheError';
+    // eslint-disable-next-line no-magic-numbers
+    this.code = 2;
```
agnivade, line 22: What is the purpose of having error codes in the class object?

lib/errors.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+module.exports = {
+  TldrError,
+  EmptyCacheError,
+  MissingPageError
+};
+
+function trim(strings, ...values) {
```
agnivade, line 44: What is the purpose of this? Is it just to remove the newlines in the original string?

lib/errors.js
vladimyr: Sure but why? It'd just pollute the global namespace without any good reason?

lib/errors.js
vladimyr: To reuse it as exit code as spec mandates different exit codes for different fatal errors.

lib/errors.js
vladimyr: Syntactical sugar to replace the following form:
```js
    super(`
Page not found.
If you want to contribute it, feel free to send a pull request to: ${repo}
    `.trim());
```
with slightly more readable IMHO:
```js
    super(trim`
Page not found.
If you want to contribute it, feel free to send a pull request to: ${repo}
    `);
```

lib/errors.js
agnivade: Ok.

agnivade (approved):

PR #352, opened 2021-05-19
add support for node v16

navarroaxel:
## Description

## Checklist

- [ ] Code compiles correctly

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [10.x, 12.x, 13.x, 14.x]
+        node-version: [12, 14, 16]
 
     steps:
     - name: Cancel Previous Runs
@@ -34,7 +34,7 @@ jobs:
           ${{ matrix.os }}-npm-
 
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node-version }}
 
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   ],
   "homepage": "http://tldr-pages.github.io",
   "engines": {
-    "node": ">=6.12.0"
+    "node": ">=12.0.0"
   },
   "main": "bin/tldr",
   "files": [


vladimyr:
I'm not eager on dropping v10 support either but I guess it can be safely done since it reached the EOL stage just recently: https://github.com/nodejs/Release/tree/14e3d74#end-of-life-releases

vladimyr (approved):

agnivade:
Should we also update the min engine version here: https://github.com/tldr-pages/tldr-node-client/blob/3511d91454e1056d4c0926216a15044295734f0c/package.json#L25 to be above 10?

vladimyr:
It should reflect tested versions, so `>=12` seems more appropriate.

bl-ue:
Ha! It went into Maintenance LTS Start on 2020-05-19, one year ago today! 😮

owenvoke (approved):

bl-ue (approved):

agnivade:
@vladimyr - Do you feel this needs a major version from the tool's perspective? Or it's okay to bump the minor version?

vladimyr:
@agnivade Take a look at this example https://github.com/sindresorhus/open-cli/releases Bump from v5 to v6 was exactly due to a change in minimum supported node version so my answer is: yes we need a major version bump.

package.json
```diff
@@ -22,7 +22,7 @@
   ],
   "homepage": "http://tldr-pages.github.io",
   "engines": {
-    "node": ">=6.12.0"
+    "node": ">=12.0.0"
```
vladimyr, line 25:
```suggestion
    "node": ">=12"
```
You don't need to fully specify a version.

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [10.x, 12.x, 13.x, 14.x]
+        node-version: [12, 14, 16]
```
vladimyr, line 16: What about v13 and v15? They are not LTS versions but we still officially support them so they should really be a part of our test matrix?

vladimyr (changes requested):

.github/workflows/test.yml
bl-ue: But isn't v13 EOL?

vladimyr: It is but I don't think there is good reason to further complicate the version support scheme unless it is really necessary. Simple min required version should be enough.

navarroaxel: We can assume that just a little segment of enthusiastic use non-LTS versions, also the probabilities to fail in v13 or v15 is too small. And we don't have the resources to maintain that kind of builds, and it's ok big project doesn't officially support non-LTS versions of Ubuntu (could work or not).

PR #355, opened 2021-05-25
add jsdocs to tldr and cache

navarroaxel:
## Description

## Checklist

- [ ] Code compiles correctly
- [ ] All tests passing (`npm run test:all`)

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -16,10 +16,19 @@ class Cache {
     this.cacheFolder = path.join(config.cache, 'cache');
   }
 
+  /**
+   * Fetch stats from the cache folder.
+   * @returns {Promise<any>} A promise with the stats of the cache folder.
+   */
   lastUpdated() {
     return fs.stat(this.cacheFolder);
   }
 
+  /**
+   * Fetch a page from cache using preferred language and preferred platform.
+   * @param page {} A
+   * @returns {Promise<unknown>}
+   */
   getPage(page) {
     let preferredPlatform = platform.getPreferredPlatformFolder(this.config);
     const preferredLanguage = process.env.LANG || 'en';
@@ -36,10 +45,18 @@ class Cache {
       });
   }
 
+  /**
+   * Clean the cache folder.
+   * @returns {Promise<any>} A promise when the remove is completed.
+   */
   clear() {
     return fs.remove(this.cacheFolder);
   }
 
+  /**
+   * Update the cache folder and returns the English entries.
+   * @returns {Promise<any>} A promise with the index.
+   */
   update() {
     // Temporary folder path: /tmp/tldr/{randomName}
     const tempFolder = path.join(os.tmpdir(), 'tldr', utils.uniqueId());
--- a/lib/index.js
+++ b/lib/index.js
@@ -91,17 +91,24 @@ function hasLang(targets, preferredLanguage) {
   });
 }
 
-// hasPage is always called after the index is created,
-// hence just return the variable in memory.
-// There is no need to re-read the index file again.
+/**
+ * Check if a page is in the index.
+ * @returns {boolean} A boolean that indicates if the page is present.
+ */
 function hasPage(page) {
+  // hasPage is always called after the index is created,
+  // hence just return the variable in memory.
+  // There is no need to re-read the index file again.
   if (!shortIndex) {
     return false;
   }
   return page in shortIndex;
 }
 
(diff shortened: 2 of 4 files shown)


blueskyson (approved):
Wow, I wonder why comment enhancements are left so long unmerged.

kbdharun (approved):

lib/cache.js
```diff
@@ -16,10 +16,19 @@ class Cache {
...
+   * @returns {Promise<any>} A promise with the stats of the cache folder.
+   */
   lastUpdated() {
     return fs.stat(this.cacheFolder);
   }
 
+  /**
+   * Fetch a page from cache using preferred language and preferred platform.
+   * @param page {} A
+   * @returns {Promise<unknown>}
```
agnivade, line 30: What is the difference between `Promise<any>` and `Promise<unknown>` ?

lib/index.js
```diff
@@ -110,6 +117,13 @@ function commands() {
 
 // Return all commands for a given platform.
 // P.S. - The platform 'common' is always included.
```
agnivade, line 119: Should we remove this then? Seems repeated.

lib/tldr.js
```diff
@@ -21,6 +21,11 @@ class Tldr {
     this.cache = new Cache(this.config);
   }
 
+  /**
+   * Print pages for a given platform..
```
agnivade, line 25: Multiple periods at the end of the sentence.

lib/tldr.js
```diff
@@ -29,17 +34,33 @@ class Tldr {
...
   listAll(singleColumn) {
     return index.commands()
       .then((commands) => {
         return this.printPages(commands, singleColumn);
       });
   }
 
+  /**
+   * Print a command page.
+   * @param commands {string[]} A given command to be printed.
```
agnivade, line 51: "A given command" is incorrect when we are passing an array. Let's change this to plural.

lib/tldr.js
```diff
@@ -82,24 +112,41 @@ class Tldr {
...
+   */
   clearCache() {
     return this.cache.clear().then(() => {
       console.log('Done');
     });
   }
 
+  /**
+   * Update the cache.
+   * @returns {Promise<any>} A promise with the index.
```
agnivade, line 127: Should this be "A promise with the cache"?

lib/tldr.js
```diff
@@ -29,17 +34,33 @@ class Tldr {
       });
   }
 
+  /**
+   * Print all pages in the cache.
```
vitorhcl, line 38:
```suggestion
   * Print the name of all pages in the index.
```

lib/tldr.js
```diff
@@ -108,6 +155,12 @@ class Tldr {
       });
   }
 
+  /**
+   * Print all pages.
```
vitorhcl, line 159:
```suggestion
   * Print the name of all pages.
```

vitorhcl (approved):
LGTM after we address the left concerns.

However, we can resolve just some of them and merge it, this PR adds documentation for quite a lot of functions. It's better to have a not ideal documentation than almost no documentation.

vitorhcl (commented):
https://github.com/tldr-pages/tldr-node-client/pull/355#discussion_r1264822592:
> What is the difference between `Promise<any>` and `Promise<unknown>` ?

See https://stackoverflow.com/questions/51439843/unknown-vs-any. In summary:
> More verbosely, unknown is I don't know (yet), thus I have to figure it out, any is I don't care, thus I don't care

I think there are too many `any`s/`unknown`s here, but we can change them later if we need.

agnivade (commented):
I still see my comments not addressed yet. Feel free to request a review once that's done. Thanks.

vitorhcl:
IMO we can take over this PR, since it was opened 2 years ago with no response since then.

lib/cache.js
SethFalco: These are both reserved keywords in `tsc`. I'm not sure if other type checking users them.

* `any` means to disable type checking outright for the result.
* `unknown` means keeps type checking enabled, but enforces that the consumer of the Promise must be very defensive if they are to access any properties in the result.

In practice, this means that if the return is `Promise<any>` they caller can just make any wild assumption about what the result is without any complaints from the type checker. Meanwhile, with `Promise<unknown>` the caller will have to verify that properties exist before it can access them, because the type is documented as unknown at runtime.

PR #366, opened 2021-11-01
Show error on failing to parse .tldrrc

### LINKED ISSUE #365: tldr fails silently when trying to read invalid `.tldrrc` leading to confusion
I come from here: https://github.com/tldr-pages/tldr/issues/1262

https://github.com/tldr-pages/tldr-node-client/blob/1641dea94df3aabcd36f381dcac61a50905a36ea/lib/config.js#L16-L21

There should be some error printed on error stream that the JSON object couldn't be loaded due to invalid syntax. JSON parse fails silently when the file is missing (that's OK), but SHOULD error when the config is invalid. Common examples when `*rc` files print errors include `.vimrc`, `.tmux.conf` and probably others.

The simplest case - yeet the ESlint directives and add proper error handling.

T3sT3ro:
## Description
Fixes #365 

## Checklist

- [x] Code compiles correctly

^^^ I am on such a bad connection right now, that I can't even do `npm install` correctly. I need someone to check, review and enhance if needed. I've never worked with mocha before, so also please check the test I added. Maybe it would also be good to add a test that checks what happens if `.tldrrc` file is missing.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -16,7 +16,7 @@ exports.get = () => {
   try {
     customConfig = JSON.parse(fs.readFileSync(CUSTOM));
   } catch (ex) {
-    if(ex instanceof SyntaxError) {
+    if (ex instanceof SyntaxError) {
       throw new Error('The content of .tldrrc is not a valid JSON object:\n' + ex);
     }
   }
--- a/lib/config.js
+++ b/lib/config.js
@@ -16,8 +16,9 @@ exports.get = () => {
   try {
     customConfig = JSON.parse(fs.readFileSync(CUSTOM));
   } catch (ex) {
-    if(ex instanceof SyntaxError)
+    if(ex instanceof SyntaxError) {
       throw new Error('The content of .tldrrc is not a valid JSON object:\n' + ex);
+    }
   }
 
   let merged = defaults(customConfig, defaultConfig);
--- a/lib/config.js
+++ b/lib/config.js
@@ -16,8 +16,8 @@ exports.get = () => {
   try {
     customConfig = JSON.parse(fs.readFileSync(CUSTOM));
   } catch (ex) {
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -16,8 +16,8 @@ exports.get = () => {
   try {
     customConfig = JSON.parse(fs.readFileSync(CUSTOM));
   } catch (ex) {
-    /*eslint-disable */
-    /*eslint-enable */
+    if(ex instanceof SyntaxError)
+      throw new Error('The content of .tldrrc is not a valid JSON object:\n' + ex);
```
marchersimon, line 20:
```suggestion
    if(ex instanceof SyntaxError) {
      throw new Error('The content of .tldrrc is not a valid JSON object:\n' + ex);
    }
```

lib/config.js
```diff
@@ -16,8 +16,9 @@ exports.get = () => {
   try {
     customConfig = JSON.parse(fs.readFileSync(CUSTOM));
   } catch (ex) {
-    /*eslint-disable */
-    /*eslint-enable */
+    if(ex instanceof SyntaxError) {
```
MasterOdin, line 19: This kind of thing should be covered by eslint...

```suggestion
    if (ex instanceof SyntaxError) {
```

MasterOdin (approved):

lib/config.js
T3sT3ro: The only thing the lint complains about are warnings about todos.

MasterOdin: Yeah, that's what I mean. We should update the earlier configuration to enforce uniform if statement usage.

owenvoke (approved):

navarroaxel (approved):

marchersimon (approved):
Thanks @T3sT3ro

PR #386, opened 2022-08-17
README: Fix variable assignment on Zsh

JakobMiksch:
## Description
Fix variable assignment on Zsh in README:


agnivade (approved):
Thanks!

PR #405, opened 2023-02-02
Add node 16 and 18 to test matrix

MasterOdin:
## Description

PR adds node 16 and 18 to our test matrix to ensure the client works against the latest LTS releases. Unlike #352, I chose not to bump the minimum engine version as the codebase given that not yet using anything that wouldn't work against node 10.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [10.x, 12.x, 13.x, 14.x]
+        node-version: [10.x, 12.x, 13.x, 14.x, 16.x, 18.x]
 
     steps:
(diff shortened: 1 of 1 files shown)


agnivade:
Node 10,12,13 have gone out of support quite a while back. IMO it's a bit wasteful to use CI resources on those versions. I think we should do a major version release and drop support for those old versions.

owenvoke:
@agnivade, that definitely makes sense too 👍🏻

MasterOdin:
I would prefer to do that in a follow-up PR as I view the two topics of "test against newer versions of node" and "raise the lowest supported node version" as separate.

agnivade:
> I would prefer to do that in a follow-up PR as I view the two topics of "test against newer versions of node" and "raise the lowest supported node version" as separate.

Both are related, but not a big deal. We can merge this.

PR #407, opened 2023-02-07
Drop support for node 12 and lower

### LINKED ISSUE #367: Bump minimum engine version to 8+
The current minimum engine version is Node 6 for this client. That version of node was EOL in April 2019. The next LTS version (8) was EOL in December 2019 and 10.x was EOL in April 2021.

While we could jump to 10+, minimally, it would be nice to move to 8+ as it would allow taking advantage of async/await for promises and greatly clean-up some of the nested promise chains the project heavily uses.

For example:

Current code:
```js
  printBestPage(command, options={}) {
    // Trying to get the page from cache first
    return this.cache.getPage(command)
      .then((content) => {
        // If found in first try, render it
        if (!content) {
          return spinningPromise('Page not found. Updating cache...', () => {
            return this.cache.update();
          })
            .then(() => {
              return spinningPromise('Creating index...', () => {
                return search.createIndex();
              });
            })
            .then(() => {
              // And then, try to check in cache again
              return this.cache.getPage(command);
            });
        }
        return content;
      })
      .then((content) => {
        if (!content) {
          throw new MissingPageError(this.config.pagesRepository);
        }
        return this.checkStale()
          .then(() => {
            this.renderContent(content, options);
          });
      });
  }
```

async/await version:
```js
  async printBestPage(command, options={}) {
    // Trying to get the page from cache first
    let content = await this.cache.getPage(command);
    if (!content) {
      await spinningPromise('Page not found. Updating cache...', () => {
        return this.cache.update();
      });
      await spinningPromise('Creating index...', () => {
        return search.createIndex();
      });
      content = await this.cache.getPage(command);
    }
    if (!content) {
      throw new MissingPageError(this.config.pagesRepository);
    }
    await this.checkStale();
    return this.renderContent(content, options);
  }
```

14 lines shorter and no nesting, and exact same functionality (I think, I eyeballed this).

There are a number of places that would benefit from this. The downside would be breaking compatibility with old versions of node, but given how EOL it's been, it should be fine, and of course old client versions would still work.

MasterOdin:
## Description

This PR drops support for versions of node less than 14, where the following versions had the following EOL:
* Node 8 was 12/31/2019
* Node 10 was 04/30/2021
* Node 12 was 04/30/2022

Node 14 is not yet EOL, though it is upcoming, but probably good to wait a bit of time past the EOL before totally dropping it.

By raising the minimum version, we can use the following new features within the codebase:
* [async/await](https://javascript.info/async-await) (available in node 8)
* [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) (available in node 12)

Closes #367 

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [10.x, 12.x, 13.x, 14.x, 16.x, 18.x]
+        node-version: [14.x, 16.x, 18.x]
 
     steps:
     - name: Cancel Previous Runs
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   ],
   "homepage": "http://tldr-pages.github.io",
   "engines": {
-    "node": ">=6.12.0"
+    "node": ">=14"
   },
   "main": "bin/tldr",
   "files": [


owenvoke (approved):

agnivade (approved):
Thanks!

PR #409, opened 2023-02-08
Move to v2 format for package-lock.json

MasterOdin:
## Description

PR modifies the `package-lock.json` file to the [new v2 format](https://blog.npmjs.org/post/621733939456933888/npm-v7-series-why-keep-package-lockjson) which was rolled out with npm v7 back in 10/2020. This is supposed to be better able to give deterministic build results with regards to tree shape. The new lockfile is backwards compatible with npm v6 (which comes with node 14 on some platforms like GH actions), but it's expected that most people will be trending forward to newer versions of npm (which is at v9), that most people won't need to worry about the compatibility.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8217 @@
 {
   "name": "tldr",
   "version": "3.3.8",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "tldr",
+      "version": "3.3.8",
+      "license": "MIT",
+      "dependencies": {
+        "axios": "^0.21.1",
+        "chalk": "^4.1.0",
+        "commander": "^6.1.0",
+        "fs-extra": "^9.0.1",
+        "glob": "^7.1.6",
+        "lodash": "^4.17.20",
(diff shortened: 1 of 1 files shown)


MasterOdin:
If someone uses npm v7+, then the format of the lockfile will remain v2.

If someone uses npm v6, then the format of the lockfile will be downgraded to v1. This is the thing we'd want to watch out for in someone submitting a PR that modifies dependencies.

To know what version the lockfile is at, there's a `lockfileVersion` field that is at the top of the file that designates the version, where if that's missing, then it was made using npm v5 or lower, if it's 1, then that's npm 6 and 2 or higher is npm v7+. If we wanted to totally break allowing people with older versions of npm, the version could be set to 3, but there's not a huge gain there, and again, in my experience most people are already using a newer version of npm anyway, and I've not seen issues of package-lock versioning once upgraded.

Moving between the versions is done as part of the regular `npm install` usage, and I "upgraded" the `package-lock.json` by doing `npm install`, though if you wanted to fully replicate my process, you could do `npx npm@8 install`

agnivade (approved):
Thanks!

PR #412, opened 2023-03-11
README: Update chatroom badge to matrix

kbdharun:
Related PR:- https://github.com/tldr-pages/tldr/pull/9898

IDK why but can't request a review from @waldyrious so doing a ping instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![NPM version][npm-image]][npm-url]
 [![GitHub Action Build Status][gh-actions-image]][gh-actions-url]
-[![Gitter chat][gitter-image]][gitter-url]
+[![Matrix chat][matrix-image]][matrix-url]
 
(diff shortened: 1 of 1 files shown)


owenvoke (approved):

waldyrious:
> IDK why but can't request a review from @waldyrious so doing a ping instead.

That's because I was removed as an org member in https://github.com/tldr-pages/tldr/issues/2257. In retrospect, that process (which, for the record, I [authored](https://github.com/tldr-pages/tldr/pull/1839) :sweat_smile:) was not the best way to handle inactive maintainers, since indeed I have been able to come back to semi-active maintenance every now and then, and not having access to the repos has hindered my ability to contribute (case in point). Indeed, I don't think any others have been in that manner, so at least the _de facto_ practice hasn't created this problem for others. Revising the policy is one of my goals for when I am able to resume active participation in the project. No promises on when that will happen, though! :sweat_smile:

PR #420, opened 2023-10-04
Fix: Download Only the files for Required Language And English as Backup | Changed the package for Unzipping

vivekjoshi556:
## Description
The task at hand was to update the client so that it would only download the zip files for languages that were supported by the user system. In any case, files in english should be downloaded to act as backup. 

## Explanation of Solution
- While doing this I realized that the unzip library being used was causing issue as it was emitting the close event on extractor twice and because of it the extraction wasn't completing and some of the files were not getting extracted. Trying to resolve it I saw similiar issues on the official repo and the issue wasn't resolved. So I decided to change the npm package to unzip.
- It uses both ll & cc to download the correct zip file whenever possible.
- Is there a way to determine which languages have different version available (like pt has pt_BR, pt_PT), right now I have hard-coded the values in util.js and it will need to be updated in case other versions for other languages are added.

## Checklist

- [x] Code compiles correctly

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,6 +1,6 @@
 {
   "pagesRepository": "https://github.com/tldr-pages/tldr",
-  "repository": "https://tldr.sh/assets/tldr.zip",
+  "repositoryBase": "https://tldr.sh/assets/tldr-pages",
   "skipUpdateWhenPageNotFound": false,
   "themes": {
     "simple": {
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -13,6 +13,16 @@ class Cache {
     // TODO: replace this with a private field when it reaches enough maturity
     // https://github.com/tc39/proposal-class-fields#private-fields
     this.config = config;
+
+    this.languages = ['en'];
+    // Get the primary & secondary language.
+    this.languages = this.languages.concat([
+      ...new Set(utils.localeToLang(process.env.LANG)),
+    ]);
+    this.languages = this.languages.concat([
+      ...new Set(utils.localeToLang(process.env.LANGUAGE)),
+    ]);
+
     this.cacheFolder = path.join(config.cache, 'cache');
   }
 
@@ -23,7 +33,8 @@ class Cache {
   getPage(page) {
     let preferredPlatform = platform.getPreferredPlatformFolder(this.config);
     const preferredLanguage = process.env.LANG || 'en';
-    return index.findPage(page, preferredPlatform, preferredLanguage)
+    return index
+      .findPage(page, preferredPlatform, preferredLanguage)
       .then((folder) => {
         if (!folder) {
           return;
@@ -45,30 +56,37 @@ class Cache {
     const tempFolder = path.join(os.tmpdir(), 'tldr', utils.uniqueId());
 
     // Downloading fresh copy
-    return Promise.all([
-      // Create new temporary folder
-      fs.ensureDir(tempFolder),
-      fs.ensureDir(this.cacheFolder)
-    ])
-      .then(() => {
-        // Download and extract cache data to temporary folder
-        return remote.download(tempFolder);
-      })
-      .then(() => {
-        // Copy data to cache folder
-        return fs.copy(tempFolder, this.cacheFolder);
-      })
-      .then(() => {
-        return Promise.all([
-          // Remove temporary folder
-          fs.remove(tempFolder),
-          index.rebuildPagesIndex()
-        ]);
-      })
-      // eslint-disable-next-line no-unused-vars
-      .then(([_, shortIndex]) => {
-        return shortIndex;
-      });
+    return (
+      Promise.all([
+        // Create new temporary folder
+        fs.ensureDir(tempFolder),
+        fs.ensureDir(this.cacheFolder),
+      ])
+        .then(() => {
+          // Download and extract cache data to temporary folder
+          let activeDownloads = [];
+          for (let lang of this.languages) {
+            activeDownloads.push(remote.download(tempFolder, lang));
+          }
+
+          return Promise.allSettled(activeDownloads);
+        })
+        .then(() => {
+          // Copy data to cache folder
+          return fs.copy(tempFolder, this.cacheFolder);
+        })
+        .then(() => {
+          return Promise.all([
+            // Remove temporary folder
+            fs.remove(tempFolder),
+            index.rebuildPagesIndex(),
+          ]);
+        })
+        // eslint-disable-next-line no-unused-vars
+        .then(([_, shortIndex]) => {
+          return shortIndex;
+        })
+    );
   }
 }
 
--- a/lib/remote.js
+++ b/lib/remote.js
@@ -1,34 +1,40 @@
 'use strict';
 
-const unzip = require('node-unzip-2');
+const path = require('path');
+const fs = require('fs-extra');
+const unzip = require('adm-zip');
 const config = require('./config');
 const axios = require('axios');
 
 // Downloads the zip file from github and extracts it to folder
-exports.download = (path) => {
-  const url = config.get().repository;
+exports.download = (loc, lang) => {
+  // If the lang is english then keep the url simple, otherwise add language.
+  const url = config.get().repositoryBase + (lang === 'en' ? '' : '.' + lang) + '.zip';
+  const folderName = path.join(loc, 'pages' + (lang === 'en' ? '' : '.' + lang));
 
-  // Creating the extractor
-  const extractor = unzip.Extract({ path });
-
-  let req = axios({
+  return axios({
     method: 'get',
     url: url,
     responseType: 'stream',
     headers: { 'User-Agent' : 'tldr-node-client' }
   }).then(function (response) {
-    response.data.pipe(extractor);
-  });
+    return new Promise((resolve, reject) => {
+      let fileName = path.join(loc, 'download_' + lang + '.zip');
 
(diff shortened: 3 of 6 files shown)


agnivade:
Can we add some tests please?

vivekjoshi556:
@agnivade Sure I'll add the tests

> Can we add some tests please?

Sure I'll add the tests. 
Can you please help me with the failing job in the workflow? All the jobs passed when I made the first commit before the PR, but now one of them is failing. It happened in the repo I forked and when I tried to run the job again alone it passed.

kbdharun:
Btw, @vivekjoshi556 

Workflow requiring approval before running is enabled for first time contributors so that maintainers can review the code before running the action. (This is to minimize multiple failed runs). After this PR is merged, in future the workflow will run by default for you (without requiring approval).

vivekjoshi556:
> Workflow requiring approval before running is enabled for first time contributors so that maintainers can review the code before running the action. (This is to minimize multiple failed runs). After this PR is merged, in future the workflow will run by default for you (without requiring approval).
> 
> >

I understand they require approval. I was talking about the last time the jobs ran for my changes.  
Actions run [link](https://github.com/tldr-pages/tldr-node-client/actions/runs/6410031720).
In this run the job for ubuntu-latest, 20.x failed. All others passed.

kbdharun:
> I understand they require approval. I was talking about the last time the jobs ran for my changes.
Actions run [link](https://github.com/tldr-pages/tldr-node-client/actions/runs/6410031720).
In this run the job for ubuntu-latest, 20.x failed. All others passed.

Oh, it is possible that the action run was cancelled as it reached 6 hours limit (default for action runs in GitHub's runners).

vivekjoshi556:
@kbdharun please check this PR & let me know if anything else needs to be updated.

lib/cache.js
```diff
@@ -45,30 +46,37 @@ class Cache {
...
+        fs.ensureDir(this.cacheFolder),
+      ])
+        .then(() => {
+          // Download and extract cache data to temporary folder
+          let activeDownloads = [];
+          for (let lang of this.config.languages) {
+            activeDownloads.push(remote.download(tempFolder, lang));
+          }
+
+          return Promise.allSettled(activeDownloads);
```
MasterOdin, line 62: Could rewrite this to be a bit more succinct by using a map I think?
```
return Promise.allSettled(this.config.languages.map((lang) => remote.download(tempFolder, lang)));
```

lib/cache.js
```diff
@@ -45,30 +46,37 @@ class Cache {
...
-          // Remove temporary folder
-          fs.remove(tempFolder),
-          index.rebuildPagesIndex()
-        ]);
-      })
-      // eslint-disable-next-line no-unused-vars
-      .then(([_, shortIndex]) => {
-        return shortIndex;
-      });
+    return (
```
MasterOdin, line 49: Do we need to reform this entire block of code, or can the previous indentation be left in place? Personally seeing a `return (` just feels weird vs having `return Promise.all([` on the one line.

lib/config.js
```diff
@@ -35,6 +36,20 @@ exports.get = () => {
   if (errors.length > 0) {
     throw new Error('Error in .tldrrc configuration:\n' + errors.join('\n'));
   }
+
+  // Setting correct languages
+  merged['languages'] = ['en'];
```
MasterOdin, line 41: Using similar conventions as above, better to do `merged.languages` vs `merged['languages']`.

lib/config.js
```diff
@@ -35,6 +36,20 @@ exports.get = () => {
...
+  // Get the primary & secondary language.
+  let langs = utils.localeToLang(process.env.LANG);
+  merged['languages'] = merged['languages'].concat(langs);
+  
+  if(process.env.LANGUAGE !== undefined) {
+    let langs = process.env.LANGUAGE.split(':');
+    for(let lang of langs) {
+      merged['languages'] = merged['languages'].concat(utils.localeToLang(lang));
+    }
```
MasterOdin, line 50: Similar as above, can use `map` to make this more succinct:

```suggestion
    merged.languages.push(...langs.map((lang) => utils.localeToLang(lang)));
```

The maximum number of languages is also low enough that doing a spread to push vs individually will have no practical difference in terms of speed/memory.

lib/remote.js
```diff
@@ -1,34 +1,43 @@
...
 const config = require('./config');
 const axios = require('axios');
 
 // Downloads the zip file from github and extracts it to folder
-exports.download = (path) => {
-  const url = config.get().repository;
+exports.download = (loc, lang) => {
+  // If the lang is english then keep the url simple, otherwise add language.
+  const url = config.get().repositoryBase + (lang === 'en' ? '' : '.' + lang) + '.zip';
+  const folderName = path.join(loc, 'pages' + (lang === 'en' ? '' : '.' + lang));
```
MasterOdin, line 13: Better to define our path suffix once and reuse it, make it easier to modify in the future:

```suggestion
  const suffix = lang === 'en' ? '' : '.' + lang;
  const url = config.get().repositoryBase + suffix + '.zip';
  const folderName = path.join(loc, 'pages' + suffix);
```

test/remote.spec.js
```diff
@@ -0,0 +1,105 @@
...
+
+    afterEach(() => {
+      // Clearing Spies & Stubs
+      fs.copy.restore();
+      fs.remove.restore();
+      fs.ensureDir.restore();
+      utils.localeToLang.restore();
+      index.rebuildPagesIndex.restore();
+      
+      fs.remove(tempFolder);
```
MasterOdin, line 88: Should probably either `await` this or use `fs.removeSync`, where if this ends up throwing for whatever reason, then since it can happen on a different tick as the `afterEach` callback, can lead to weird test failures that are hard to debug.

test/remote.spec.js
```diff
@@ -0,0 +1,105 @@
...
+        expectedFolders: ['pages', 'pages.hi', 'pages.pt_BR'],
+      },
+    ];
+
+    beforeEach(() => {
+      // Iterate over the test cases
+      sinon.spy(fs, 'ensureDir');
+      sinon.stub(fs, 'remove').resolves();
+      sinon.stub(fs, 'copy').resolves();
+      sinon.stub(utils, 'localeToLang').returns(testCases[i].LANG);
```
MasterOdin, line 51: Instead of keeping track of an index variable to use in a `beforeEach`, it's more straightforward I think to do something like this:

```
testCases.forEach((testCase) => {
  describe(testCase.description, () => {
    before(() => {
      // set up sinon
    });
    it('passes', () => {
      // test
    });
    after(() => {
      // break down sinon
    });
  });
});
```

config.json
```diff
@@ -40,4 +40,4 @@
     }
   },
   "theme": "simple"
-}
+}
```
MasterOdin, line 43: I wouldn't remove the ending newline here.

package.json
```diff
@@ -80,4 +80,4 @@
     "type": "liberapay",
     "url": "https://liberapay.com/tldr-pages"
   }
-}
+}
```
MasterOdin, line 83: I wouldn't remove the ending newline here.

lib/cache.js
vivekjoshi556: Looks great. I'll update the file.

Note: There is a linting rule "arrow-parens" problem with this, so I'll just include extra parentheses and update the code.

test/remote.spec.js
vivekjoshi556: Thank you for this. It was my first time using mocha so I was struggling with this part a bit.

vivekjoshi556:
@MasterOdin I am done with the mentioned changes. Please review the PR.

lib/cache.js
MasterOdin: Looks like this got missed. I went ahead and made this change in 04a8cf2414edb0c82ef02f29a355c9f4b27dcccf.

lib/remote.js
```diff
@@ -1,34 +1,44 @@
...
 
 // Downloads the zip file from github and extracts it to folder
-exports.download = (path) => {
-  const url = config.get().repository;
+exports.download = (loc, lang) => {
+  // If the lang is english then keep the url simple, otherwise add language.
+  const suffix = (lang === 'en' ? '' : '.' + lang);
+  const url = config.get().repositoryBase + suffix + '.zip';
+  const folderName = path.join(loc, 'pages' + suffix);
+  const REQUEST_TIMEOUT = 10000;
```
kbdharun, line 15: Regarding this, we recently added support for the English archive in the same format `tldr-pages-en.zip` (after a request from tealdeer's maintainer). Can it be implemented here instead?

vivekjoshi556: sure should be easy.

MasterOdin: It should be noted that only the archive name should be generalized here, while still leaving in place that the english archive is extracted to the `pages` directory. Changing that behavior is a much larger one, which should be done in a separate PR that's focused purely on that, as will also have to take into account backwards compatibility.

vivekjoshi556: So, won't it be better to handle this entirely in the same (other) PR?

MasterOdin: I'd say so, but not sure how much @kbdharun wants to see the archive name changed here. The old zip will continue to work anyway for the time being that it's not exactly urgent to make that change either.

kbdharun (approved):

kbdharun:
I guess this PR can be merged first, then I will rebase mine after this.

Issue #353, opened 2021-05-19
Improve update cache process


navarroaxel:
## Actual behavior
Updating this client’s cache takes ~41 seconds.
Loading the pages' cache takes `< 1 second`.
The cache for the `--search` argument takes the rest of the time.

### Explaining the build index process
The process reads every page building a dictionary with `{key: count}` with `key` as the word and `count` with the count of its appearance in the given page.

### Options to improve this process

1. Create the dictionary when `--search` is executed and the `search-corpus.json` file isn't created. Then clean the file when the index is updated.
2. Split the process into threads (child process, threads or whatever) using chunks or one thread per platform.
3. Use a background process to create the `search-corpus.json`.
4. Use the solution proposed in #350 (don't update the dictionary on every miss).

## Environment
- Operating system - Arch Linux
- Node.js version any (tested in v12 and v16)

sbrl:
It's also possible that there are some performance gains to be had in the search index calculation process, but I haven't looked too closely at it to know.

agnivade:
I'd prefer 2 or 3. Or even take a deeper look into the search index build process and look if there are any low hanging fruits there as @sbrl says. The search index was a one-off feature added by a contributor and wasn't really touched after that. So I won't be surprised if there are some simple gains to reap.

vladimyr:
Honestly, building a search index on the end-user machine is wasting resources. Each time cache is updated (tarball downloaded and extracted) it needs to get rebuilt so why don't we simply prebuild it and ship it together with pages in the first place?

agnivade:
I like that. I'd also prefer to have a default build and a minimal build with no search index for users who don't use search at all. And for this one, we can dynamically build the index if the user decides to do a search at some point.

vladimyr:
> And for this one, we can dynamically build the index if the user decides to do a search at some point.

Or simply download it at page search time.

navarroaxel:
I don't think so, each client should use the data structure that it finds as a better fit for the language and platform. Also, that sounds like a hard dependency that I don't see compatible with this project.

vladimyr:
> I don't think so, each client should use the data structure that it finds as a better fit for the language and platform.

Um, I don't see how prebuilding search index for node client using GH actions changes that?

> Also, that sounds like a hard dependency that I don't see compatible with this project.

What kind of dependency we are talking about?

navarroaxel:
> Um, I don't see how prebuilding search index for node client using GH actions changes that?

[...]

Issue #350, opened 2021-05-06
Feature Request: Wait 30 minutes to update cache again


qpwo:
## Desired behaviour

```
$ tldr doesnotexist
✔ Page not found. Updating cache...
✔ Creating index...
... # takes 25 seconds for me
$ tldr doesnotexist # run again immediately
✔ Page not found. Cache updated 2 minutes ago. Run `tldr --update` to force update.
... # 28 minutes later
$ tldr doesnotexist
✔ Page not found. Updating cache...
✔ Creating index...
```

## Current behaviour

```
$ tldr doesnotexist
✔ Page not found. Updating cache...
✔ Creating index...
...
$ tldr doesnotexist 
⠇ Page not found. Updating cache...^C
```

## Environment

```
$ sw_vers
ProductName:	macOS
ProductVersion:	11.1
BuildVersion:	20C69
$ tldr --version
3.3.7
```

agnivade:
I think it's rare to see this in a practical scenario. You want to lookup a command, you see that it doesn't exist. Would you want to retry that command again in 2 minutes? That seems unlikely. But I agree that there's definitely things to improve on. Let's track those here: https://github.com/tldr-pages/tldr-node-client/issues/353

PR #433, opened 2024-01-04
perf: improve creating tfIdf with cache

vivekjoshi556:
## Description
 Improve #353 

While building indexes there is a call to creating TfIdf.
I noticed that the the getIdf() method only depends on the word and doesn't the other values remained constant, so I cached the result of getIdf() for different words and before calling it again I check if I already had the value.

This improved the performance significantly. 

getTfIdf() performance:
Before: 1:52.468s. (maybe my sys is slow)
After: 9.021s

## Checklist

- [x ] Code compiles correctly

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -88,11 +88,15 @@ let getIdf = (word) => {
 
 let createTfIdf = () => {
   corpus.tfidf = {};
+  let idfCache = {};
+
   Object.keys(corpus.fileWords).forEach((file) => {
     corpus.tfidf[file] = {};
     Object.keys(corpus.fileWords[file]).forEach((word) => {
-      let tfidf = getTf(word, file) * getIdf(word);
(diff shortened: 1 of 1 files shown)


vivekjoshi556:
As mentioned in #350 . I know @agnivade you said

> I think it's rare to see this in a practical scenario. You want to lookup a command, you see that it doesn't exist. Would you want to retry that command again in 2 minutes? That seems unlikely. But I agree that there's definitely things to improve on. Let's track those here: https://github.com/tldr-pages/tldr-node-client/issues/353

I agree that same command might not be looked up immediately but what if user looks for some other command that also doesn't exist. 
If you agree I would like to work in this idea as well and add this suggestion in this commit as well.

kbdharun (approved):
LGTM, Thanks for this enhancement. Tested it on Windows and Linux the indexing in the update process is much faster now.

